PR #3, opened 2025-07-24
Add Makefile to generate python files based on idl definition

timl3136:
## Summary

Adds a Makefile that automates generation of Python protobuf files from Cadence's IDL definitions. Processes both API and Admin protobuf files from `idls/proto/uber/cadence/` and generates corresponding Python `_pb2.py` files in `.gen/` directory.

## Changes

- **`Makefile`** - Build automation with targets: `all` (default), `clean`, `help`, `show`
- **20 generated Python files** in `.gen/uber/cadence/api/v1/` and `.gen/uber/cadence/admin/v1/`

## Usage

```bash
make        # Generate all Python proto files
make clean  # Remove generated files
make help   # Show available targets
```

## Prerequisites

- `protoc` (Protocol Buffers compiler) in PATH
- Source `.proto` files in `idls/proto/uber/cadence/`


Makefile
```diff
@@ -0,0 +1,70 @@
...
+$(DST_DIR)/uber/cadence/api/v1/:
+	mkdir -p $(DST_DIR)/uber/cadence/api/v1
+
+$(DST_DIR)/uber/cadence/admin/v1/:
+	mkdir -p $(DST_DIR)/uber/cadence/admin/v1
+
+# Generate Python files for API protos
+$(DST_DIR)/uber/cadence/api/v1/%_pb2.py: $(PROTO_ROOT)/api/v1/%.proto | $(DST_DIR)/uber/cadence/api/v1/
+	@echo "Generating Python file for $(PROTO_ROOT)/api/v1/$*.proto"
+	protoc --proto_path=$(SRC_DIR) --python_out=$(DST_DIR) $(PROTO_ROOT)/api/v1/$*.proto
```
Groxx, line 32: we probably want to copy the protoc download-and-init-in-temp-build-dir thing from the other cadence repos :\  protoc's code is far from stable across versions.

Groxx:
copying from side-chat:
- yea seems reasonable, happy to approve if it'll streamline anything.
- versioned-protoc seems probably critical eventually.  it's a really big pain as time passes.
- .pyi files seem like a good idea to include
- it's currently written with every file doing a separate `protoc` run.  in other languages that tends to cause problems if you `make protoc -j` to run them in parallel, because it has a nasty habit of clobbering other files / reading partially-written files and getting confused.
  - is that an issue in python?  if no, the setup now is ideal.  if yes, we should probably pull it into a single protoc-target that runs them all in a specific way.

Makefile
timl3136: protoc v3.14.0 does not have pyi support, should I just a separate version for the pyi files?

PR #5, opened 2025-07-29
Set up PyProject and generated files using protoc

timl3136:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


scripts/generate_protobuf_final.py
natemort, line 1: Nit: maybe just generate_proto.py

pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: Apache Software License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.11",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+    "Topic :: System :: Distributed Computing",
+]
+requires-python = ">=3.11,<3.12"
+dependencies = [
```
natemort, line 26: Are these all used?

pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+Homepage = "https://github.com/cadence-workflow/cadence-python-client"
+Documentation = "https://cadence-python-client.readthedocs.io/"
+Repository = "https://github.com/cadence-workflow/cadence-python-client"
+"Bug Tracker" = "https://github.com/cadence-workflow/cadence-python-client/issues"
+
+[project.scripts]
+cadence-cli = "cadence.cli:main"
+
+[tool.setuptools]
+packages = ["cadence", "cadence.shared"]
```
natemort, line 66: cadence.shared is outdated now

pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+
+[tool.setuptools.package-data]
+"*" = ["*.proto", "*.thrift"]
```
natemort, line 69: I don't know what this does but I hope we don't need thrift.

requirements.txt
natemort, line 1: How does this file compare to the dependencies listed in the pyproject.toml?

scripts/generate_protobuf_final.py
```diff
@@ -0,0 +1,425 @@
...
+import sys
+import os
+from pathlib import Path
+import shutil
+import platform
+import urllib.request
+import zipfile
+
+
+def download_protoc_29_1(project_root: Path) -> str:
```
natemort, line 16: If we used the `grpc_tools.protoc` python package this might simplify the logic a bit and be more portable. What do you think? Have you explored that much?

natemort (approved):

scripts/generate_protobuf_final.py
timl3136: I tried using `grpc_tools.protoc`, but it always generate the python file with `_pb2` and `_pb2_grpc` at the end of names.

pyproject.toml
timl3136: cursor added that since idl have thrift files. But we can remove it.

requirements.txt
timl3136: We can merge then later and streamline the set up process

pyproject.toml
timl3136: I plan to trim the dependency later since it's quite a demanding operation

PR #14, opened 2025-08-26
Implement workflow & activity registry

timl3136:
**What changed?**
Implement workflow & activity registry to run workflow/activity

**Why?**
We need the registry to register workflow/activity to server in order to run it.

**How did you test it?**
unit test pass: `uv run pytest tests/cadence/worker/test_registry.py -v`

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/worker/_registry.py
+++ b/cadence/worker/_registry.py
@@ -7,25 +7,22 @@ similar to the Go client's registry.go implementation.
 """
 
 import logging
-from typing import Callable, Dict, Optional
-from dataclasses import dataclass
+from typing import Callable, Dict, Optional, Unpack, TypedDict
 
 
 logger = logging.getLogger(__name__)
 
 
-@dataclass
-class RegisterWorkflowOptions:
+class RegisterWorkflowOptions(TypedDict, total=False):
     """Options for registering a workflow."""
-    name: Optional[str] = None
-    alias: Optional[str] = None
+    name: Optional[str]
+    alias: Optional[str]
 
 
-@dataclass
-class RegisterActivityOptions:
+class RegisterActivityOptions(TypedDict, total=False):
     """Options for registering an activity."""
-    name: Optional[str] = None
-    alias: Optional[str] = None
+    name: Optional[str]
+    alias: Optional[str]
 
 
 class Registry:
@@ -46,7 +43,7 @@ class Registry:
     def workflow(
         self,
         func: Optional[Callable] = None,
-        **kwargs: RegisterWorkflowOptions
+        **kwargs: Unpack[RegisterWorkflowOptions]
     ) -> Callable:
         """
         Register a workflow function.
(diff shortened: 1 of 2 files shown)


cadence/worker/_registry.py
```diff
@@ -0,0 +1,515 @@
...
+"""
+
+import logging
+from typing import Callable, Dict, List, Optional
+
+
+logger = logging.getLogger(__name__)
+
+
+class RegistryError(Exception):
```
natemort, line 16: Do we need all these or is there an existing exception we can return instead, like a KeyError?

cadence/worker/_registry.py
```diff
@@ -0,0 +1,515 @@
...
+        self,
+        func: Optional[Callable] = None,
+        *,
+        name: Optional[str] = None,
+        alias: Optional[str] = None
+    ) -> Callable:
+        """
+        Register a workflow function.
+        
+        This method can be used as a decorator or called directly.
```
natemort, line 67: So is the idea that a user can do either:
```
registry.register_workflow(...)
```
or 
```
@registry.register_workflow(...)
def foo(): ...
```
?

That approach is pretty similar to DBOS. I think if that's the direction we want to go then I'd prefer the function names be `workflow` and `activity`, so that they're `@registry.workflow`.

cadence/worker/_registry.py
```diff
@@ -0,0 +1,515 @@
...
+        # Try the next registry in the chain if available
+        if self._next_registry:
+            try:
+                return self._next_registry.get_activity(name)
+            except ActivityNotFoundError:
+                pass
+        
+        raise ActivityNotFoundError(f"Activity '{name}' not found in registry")
+    
+    def list_workflows(self) -> List[str]:
```
natemort, line 209: I would remove all methods after (and including) this one. We should minimize the public interface.

Users are unlikely to interact heavily with the registry beyond registering activities/workflows, and these methods seem like they're of limited value. I don't expect users to need the list of activity names. We don't need to support `unregister` at all. I think we can consider adding them in later.

cadence/worker/_registry.py
```diff
@@ -0,0 +1,515 @@
...
+        Check if this registry has a next registry in the chain.
+        
+        Returns:
+            True if there is a next registry, False otherwise
+        """
+        return self._next_registry is not None
+
+
+# Global registry instance
+registry = Registry()
```
natemort, line 418: As discussed offline, we don't want a global registry.

cadence/worker/_registry.py
timl3136: discussed offline, will remove unused util functions until such specific need arises.

cadence/worker/_registry.py
timl3136: Yes, both options work and user can specify registerWorkflow/ActivityOptions if they want in both options as well.
I check this behavior in the test file

tests/cadence/worker/test_registry.py
```diff
@@ -0,0 +1,206 @@
...
+"""
+Tests for the registry functionality.
+"""
+
+import pytest
+import sys
+import os
+
+# Add the project root to the path
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))))
```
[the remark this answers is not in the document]
timl3136: yep, we don't need that thanks to uv and pytest

PR #19, opened 2025-08-28
use grpc_tools to generate proto files

shijiesheng:
**What changed?**

* use grpc_tools to generate proto files

**Why?**

avoid local compiling protoc

**How did you test it?**

Unit Test

Manual Test with client_example.py

**Potential risks**

**Release notes**

**Documentation Changes**


timl3136:
we might need to add a line in readme that in order to generate the files, it requires using dev dependency via `uv sync --extra dev` since `uv sync` will not install all the necessary modules for generate proto

timl3136 (approved):

PR #20, opened 2025-08-29
Add initial ActivityExecutor

natemort:
Create a basic activity executor capable of running functions as activities. Async functions are run as part of the main event loop, while sync ones are run in a threaadpool executor.

Type hints are expected for any parameters the activity takes, with non-typed parameters receiving the deserialized json value as a standard python type (dict, list, string, etc.).

Plenty of work remains for full activity support, such as maintaining an activity context for heartbeating.

**What changed?**
- Added basic support for an activity executor

**Why?**
- To continue work on an activity worker

**How did you test it?**
- Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tests/cadence/worker/test_worker.py
+++ b/tests/cadence/worker/test_worker.py
@@ -7,7 +7,7 @@ from unittest.mock import AsyncMock, Mock, PropertyMock
 from cadence.api.v1.service_worker_pb2 import PollForDecisionTaskRequest, PollForActivityTaskRequest
 from cadence.api.v1.tasklist_pb2 import TaskList, TaskListKind
 from cadence.client import Client
-from cadence.worker import Worker
+from cadence.worker import Worker, Registry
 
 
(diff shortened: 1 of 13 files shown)


tests/cadence/worker/test_worker.py
```diff
@@ -29,7 +29,7 @@ async def poll(_, timeout=0.0):
     type(client).domain = PropertyMock(return_value="domain")
     type(client).identity = PropertyMock(return_value="identity")
 
-    worker = Worker(client, "task_list", activity_task_pollers=1, decision_task_pollers=1, identity="identity")
+    worker = Worker(client, "task_list", Registry(), activity_task_pollers=1, decision_task_pollers=1, identity="identity")
```
timl3136, line 32: nit: should we do some assertion on the field?

timl3136 (approved):
LGTM

PR #21, opened 2025-08-29
Abstract Channel Initialization behind Client

natemort:
Our server being yarpc means that there are mandatory headers needed on every request. To include these we need to add mandatory interceptors. For additional things like retries and error mapping we'll also want additional interceptors.

GRPC's sync implementation allows for adding interceptors to an existing channel, while the async implementation does not. As a result, our client needs to be responsible for Channel creation.

Add GRPC channel options to ClientOptions and create the Channel within Client. This approach largely matches how the Java client approaches it, although it does allow for overriding the Channel still.

**What changed?**
- Abstract Channel creation behind Client

**Why?**
- So we can control the interceptors added to the Channel

**How did you test it?**
- Unit tests, and connecting to local server

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/client.py
+++ b/cadence/client.py
@@ -1,21 +1,40 @@
 import os
 import socket
-from typing import TypedDict
+from typing import TypedDict, Unpack, Any, cast
 
+from grpc import ChannelCredentials, Compression
+
+from cadence._internal.rpc.yarpc import YarpcMetadataInterceptor
 from cadence.api.v1.service_worker_pb2_grpc import WorkerAPIStub
-from grpc.aio import Channel
+from grpc.aio import Channel, ClientInterceptor, secure_channel, insecure_channel
 
 
 class ClientOptions(TypedDict, total=False):
     domain: str
+    target: str
     identity: str
+    service_name: str
+    caller_name: str
+    channel_arguments: dict[str, Any]
+    credentials: ChannelCredentials | None
+    compression: Compression
+    interceptors: list[ClientInterceptor]
+
+_DEFAULT_OPTIONS: ClientOptions = {
+    "identity": f"{os.getpid()}@{socket.gethostname()}",
+    "service_name": "cadence-frontend",
+    "caller_name": "cadence-client",
+    "channel_arguments": {},
+    "credentials": None,
+    "compression": Compression.NoCompression,
+    "interceptors": [],
+}
 
 class Client:
-    def __init__(self, channel: Channel, options: ClientOptions) -> None:
-        self._channel = channel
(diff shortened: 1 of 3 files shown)


cadence/client.py
```diff
@@ -1,21 +1,40 @@
...
+    target: str
     identity: str
+    service_name: str
+    caller_name: str
+    channel_arguments: dict[str, Any]
+    credentials: ChannelCredentials | None
+    compression: Compression
+    interceptors: list[ClientInterceptor]
+
+_DEFAULT_OPTIONS: ClientOptions = {
```
shijiesheng, line 23: maybe add target for our grpc docker socket address

cadence/client.py
```diff
@@ -1,21 +1,40 @@
...
+
+from cadence._internal.rpc.yarpc import YarpcMetadataInterceptor
 from cadence.api.v1.service_worker_pb2_grpc import WorkerAPIStub
-from grpc.aio import Channel
+from grpc.aio import Channel, ClientInterceptor, secure_channel, insecure_channel
 
 
 class ClientOptions(TypedDict, total=False):
     domain: str
+    target: str
```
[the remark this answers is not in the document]
natemort: Yeah, target is rather weird. I picked this name because it matches the naming from the GRPC api: https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.secure_channel .

In the Java client we specify host and port, I don't know the equivalent in Go.

Internally I think this will work similar to Java, where we use the GRPC transport.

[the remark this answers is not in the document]
shijiesheng: How about adding some comment on this or maybe change to grpc_target then?

cadence/client.py
```diff
@@ -24,14 +43,35 @@ def domain(self) -> str:
...
 
     @property
     def worker_stub(self) -> WorkerAPIStub:
         return self._worker_stub
 
-
     async def close(self) -> None:
         await self._channel.close()
 
+def _validate_and_copy_defaults(options: ClientOptions) -> ClientOptions:
```
shijiesheng, line 55: nit: move to ClientOptions class method so it's less likely someone implements something similar in other packages

cadence/client.py
```diff
@@ -24,14 +43,35 @@ def domain(self) -> str:
...
+    if "target" not in options:
```
shijiesheng, line 56: maybe `options.target is not None` so it's discoverable on how this field is used through IDE?

shijiesheng (approved):

timl3136 (approved):

PR #23, opened 2025-09-04
Add basic integration tests

natemort:
Use pytest-docker to run docker-compose as a pytest fixture, bringing up Cadence, Cassandra, and statsd. If possible I'd like to switch to using SQLite in the future, but I need to explore that more. For the moment this is similar to what the Java and Go clients do.

Unlike the Java and Go clients we orchestrate docker via the testing framework, rather than running the tests themselves via docker. This seems much easier to use and debug. With a bit of tinkering we should have a lot more flexibility to do things like running tests against an already running set of services to speed up iteration time.

Additionally change the pytest `asyncio_mode` to `auto` so that async tests and fixtures are automatically run in an eventloop. One major issue with the current version of pytest-asyncio that we're using is its limited control over the eventloop used for running tests/fixtures. It's currently running each in their own eventloop, which means you can't pass certain objects between them, such as GRPC's async channel objects.

**What changed?**
- Add basic integration test

**Why?**
- To prepare for more interesting ones

**How did you test it?**
- Running it locally via `uv run pytest --integration-tests`

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -73,4 +73,24 @@ jobs:
           uv sync --extra dev
       - name: Run unit tests
         run: |
-          uv run python -m pytest tests/ -v
+          uv run pytest -v
+
+  integration_test:
+    name: Integration Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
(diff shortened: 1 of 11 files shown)


shijiesheng (approved):
I learned a lot from this PR on how pytest works. Thank you!

PR #29, opened 2025-09-15
Introduce ActivityDefinition as a wrapper for Activity Fns

natemort:
In order to use activity functions within a Workflow we need to have a very explicit name for the function that we can determine without the registry, as the activity may not even be present within the same registry as the workflow. Remove the concept of aliases, and mandate a decorator on all Activity functions.

For ease of use, continue to support the decorator via the registry to both register and decorate the activity at the same time.

In addition, rather than simply storing the activity function or setting attributes on it to track that it's registered, introduce a wrapping Callable. Within the context of a Workflow we can reinterpret invocations of the activity function as the execution of the activity itself.

**What changed?**
- Require activity functions to use a decorator so that they're always explicitly named
- Add support for Activities being defined on a Class or interface
- Add support for combining Registry objects
- Wrap Activity functions so that they can be used within Workflows

**Why?**
- Better support Activity usage in workflows

**How did you test it?**
- Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


cadence/activity.py
```diff
@@ -59,3 +64,99 @@ def is_set() -> bool:
...
         return ActivityContext._var.get()
+
+
+@dataclass(frozen=True)
+class ActivityParameter:
+    name: str
+    type_hint: Type | None
+    default_value: Any | None
+
+class ExecutionStrategy(Enum):
```
timl3136, line 75: nit: naming, are we going to have more than 2 types of enum here? if not, perhaps a naming of isAsync might be more descriptive here.

cadence/activity.py
```diff
@@ -59,3 +64,99 @@ def is_set() -> bool:
...
+        return self._params
+
+    @staticmethod
+    def wrap(fn: Callable[P, T], opts: ActivityDefinitionOptions) -> 'ActivityDefinition[P, T]':
+        name = fn.__qualname__
+        if "name" in opts and opts["name"]:
+            name = opts["name"]
+
+        strategy = ExecutionStrategy.THREAD_POOL
+        if inspect.iscoroutinefunction(fn) or inspect.iscoroutinefunction(fn.__call__): # type: ignore
```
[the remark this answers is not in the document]
natemort: It covers some really weird edge cases. This PR to CPython actually has good examples: https://github.com/python/cpython/pull/99247/files

```
        class Cl:
            async def __call__(self):
                pass

        self.assertFalse(inspect.iscoroutinefunction(Cl))
        # instances with async def __call__ are NOT recognised.
        self.assertFalse(inspect.iscoroutinefunction(Cl()))
```

I don't really know why it works this way, but it does. The only likely scenario for someone to run into an issue with this is if they have other decorators on the function.

cadence/activity.py
natemort: I think there's a chance we might add multiprocessing support in the future.

abhishekj720 (approved):

PR #33, opened 2025-09-26
Define metrics with Prometheus integration

timl3136:
**What changed?**
- Define metrics using Prometheus
- also add cursorrules for running uv command

**Why?**
It enables metrics collection for monitoring Cadence workflows and activities

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/visibility/metrics.py
+++ b/cadence/_internal/visibility/metrics.py
@@ -1,0 +1,144 @@
+"""Core metrics collection interface and registry for Cadence client."""
+
+import logging
+from enum import Enum
+from typing import Dict, Optional, Protocol, Set
+
+logger = logging.getLogger(__name__)
+
+
+class MetricType(Enum):
+    """Types of metrics that can be collected."""
+
+    COUNTER = "counter"
+    GAUGE = "gauge"
+    HISTOGRAM = "histogram"
+    SUMMARY = "summary"
+
+
+class MetricsCollector(Protocol):
+    """Protocol for metrics collection backends."""
+
+    def counter(
+        self, key: str, n: int = 1, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        """Send a counter metric."""
+        ...
+
+    def gauge(
+        self, key: str, value: float, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        """Send a gauge metric."""
+        ...
+
+    def timer(
+        self, key: str, duration: float, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        """Send a timer metric."""
+        ...
+
+    def histogram(
+        self, key: str, value: float, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        """Send a histogram metric."""
+        ...
+
+
+class NoOpMetricsCollector:
+    """No-op metrics collector that discards all metrics."""
+
+    def counter(
+        self, key: str, n: int = 1, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        pass
+
+    def gauge(
+        self, key: str, value: float, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        pass
+
+    def timer(
+        self, key: str, duration: float, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        pass
+
+    def histogram(
+        self, key: str, value: float, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        pass
+
+
+class MetricsRegistry:
+    """Registry for managing metrics collection in the Cadence client."""
+
+    def __init__(self, collector: Optional[MetricsCollector] = None):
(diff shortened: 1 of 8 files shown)


cadence/_internal/visibility/prometheus.py
```diff
@@ -0,0 +1,295 @@
...
+            try:
+                self._http_server.shutdown()
+                self._http_server = None
+                logger.info("Prometheus HTTP server shutdown")
+            except Exception as e:
+                logger.error(f"Failed to shutdown HTTP server: {e}")
+
+
+# Default Cadence metrics names
+class CadenceMetrics:
```
natemort, line 270: I don't think these should be in a class, maybe an enum if you want to namespace them.

Do these match the Java/Go client or are these just examples for the sake of testing?

cadence/_internal/visibility/metrics.py
```diff
@@ -0,0 +1,86 @@
...
+        pass
+
+    def histogram(
+        self, key: str, value: float, tags: Optional[Dict[str, str]] = None
+    ) -> None:
+        pass
+
+
+# Global default handler
+_default_handler: Optional[MetricsHandler] = None
```
natemort, line 72: I don't think we should have a global for this, I'd rather we pass it into the Client which propagates to the Workers.

cadence/_internal/visibility/prometheus.py
```diff
@@ -0,0 +1,295 @@
...
+    push_to_gateway,
+    start_http_server,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class PrometheusConfig:
```
natemort, line 23: I'm not very familiar with Prometheus. Is the expectation that the user needs to configure all of this? How does the setup compare to the Java/Go clients and the server? They also all support Prometheus, right?

cadence/_internal/visibility/prometheus.py
```diff
@@ -0,0 +1,295 @@
...
+
+            if merged_tags:
+                histogram.labels(**merged_tags).observe(value)
+            else:
+                histogram.observe(value)
+
+        except Exception as e:
+            logger.error(f"Failed to send histogram {key}: {e}")
+
+    def start_http_server(self) -> None:
```
natemort, line 211: Does the user need to call this and manage the lifecycle of this object?

cadence/_internal/visibility/prometheus.py
```diff
@@ -0,0 +1,295 @@
...
+    metric_prefix: str = "cadence_"
+
+    # Default labels to apply to all metrics
+    default_labels: Dict[str, str] = field(default_factory=dict)
+
+    # Custom registry (if None, uses default global registry)
+    registry: Optional[CollectorRegistry] = None
+
+
+class PrometheusMetrics:
```
natemort, line 46: Nit: It's not necessary, but extending MetricsHandler would convey additional intent.

cadence/_internal/visibility/prometheus.py
timl3136: Thanks, I replaced them with enums.

cadence/_internal/visibility/prometheus.py
timl3136: No, I change the behavior so that it will initialize when client start

cadence/_internal/visibility/prometheus.py
```diff
@@ -0,0 +1,198 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class PrometheusConfig:
+    """Configuration for Prometheus metrics."""
+
+    # Metric name prefix
+    metric_prefix: str = "cadence_"
```
natemort, line 28: I don't think we actually want to expose this as an option unless our other clients do. It makes standard dashboards impossible.

timl3136: I will make this into a hardcoded constant to match other clients

cadence/metrics/metrics.py
```diff
@@ -0,0 +1,60 @@
...
+
+class MetricType(Enum):
+    """Types of metrics that can be collected."""
+
+    COUNTER = "counter"
+    GAUGE = "gauge"
+    HISTOGRAM = "histogram"
+
+
+class MetricsEmitter(Protocol):
```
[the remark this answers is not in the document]
timl3136: Yes, user will be using this in their workflow/activity

cadence/metrics/prometheus.py
```diff
@@ -0,0 +1,295 @@
...
+    push_to_gateway,
+    start_http_server,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class PrometheusConfig:
```
timl3136: The setup should be done at service start up, which is the same process if we want to use prometheus in java client

PR #34, opened 2025-09-29
Align DataConverter with Go Client

natemort:
**What changed?**
- Make DataConverter match Go rather than Java.

**Why?**
- The Go DataConverter's approach is unique and potentially difficult to parse. But it's unambiguous. The Java client either wraps the input into a json array, or encodes it as-is if there's a single value. This ends up being ambiguous as to how it should be decoded any time a List is one of the arguments.
- We could alternatively always wrap the arguments into a list, but introducing a third standard seems worse.

**How did you test it?**
- Unit tests

**Potential risks**
- None
- 

**Release notes**

**Documentation Changes**


timl3136 (approved):
LGTM

PR #66, opened 2026-02-05
feat: define python client search attribute types

timl3136:
**What changed?**
defined the search attribute types in python client

**Why?**
in order to use search attribute in client, we need proper typing for search attributes.

**How did you test it?**
Unit test, it's not integrated with client yet.

**Potential risks**

**Release notes**

**Documentation Changes**


shijiesheng:
implemented here https://github.com/cadence-workflow/cadence-python-client/pull/166

PR #83, opened 2026-04-01
feat: implement heartbeat progress recovery through heartbeat details

timl3136:
**What changed?**
implement heartbeat progress recovery through heartbeat details. While activity can fetch heartbeat details, it cannot yet do that during retry since retry handling is missing.

**Why?**
Allow the activity to pickup the progress that it heartbeated during retry

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/activity/_heartbeat.py
+++ b/cadence/_internal/activity/_heartbeat.py
@@ -37,5 +37,6 @@ class _HeartbeatSender:
                     identity=self._identity,
                 )
             )
+            self._previous_details = payload
         except Exception:
             _logger.warning("Heartbeat failed", exc_info=True)
--- a/cadence/_internal/activity/_heartbeat.py
+++ b/cadence/_internal/activity/_heartbeat.py
@@ -1,6 +1,7 @@
 from logging import getLogger
-from typing import Any
+from typing import Any, Type
 
+from cadence.api.v1.common_pb2 import Payload
 from cadence.api.v1.service_worker_pb2 import RecordActivityTaskHeartbeatRequest
 from cadence.api.v1.service_worker_pb2_grpc import WorkerAPIStub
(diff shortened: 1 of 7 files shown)


cadence/data_converter.py
```diff
@@ -51,6 +51,8 @@ def _decode_whitespace_delimited(
     def _convert_into(
         values: List[Any], type_hints: Sequence[Type | None]
     ) -> List[Any]:
+        if not type_hints:
```
shijiesheng: maybe it's better to do validation on from_data directly. If type_hints is None, assume there is one data inside the payload. So type_hints should default to [None]

cadence/data_converter.py
```diff
@@ -39,7 +39,7 @@ def _decode_whitespace_delimited(
     ) -> List[Any]:
         results: List[Any] = []
         start, end = 0, len(payload)
-        while start < end and len(results) < len(type_hints):
+        while start < end and (not type_hints or len(results) < len(type_hints)):
```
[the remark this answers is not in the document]
shijiesheng: move validation outside to from_data instead

cadence/_internal/activity/_heartbeat.py
```diff
@@ -15,11 +16,16 @@ def __init__(
         data_converter: DataConverter,
         task_token: bytes,
         identity: str,
+        previous_details: Payload,
     ):
         self._worker_stub = worker_stub
         self._data_converter = data_converter
         self._task_token = task_token
         self._identity = identity
+        self._previous_details = previous_details
```
shijiesheng, line 25: _previous_details need to be updated in send_heartbeat() after `await self._worker_stub.RecordActivityTaskHeartbeat`

shijiesheng (approved):

PR #84, opened 2026-04-09
feat: implement handle signal

timl3136:
**What changed?**
Added inbound signal handling. Workflows can now receive signals via @workflow.signal_method and block until a condition is met with workflow.wait_condition(predicate).

**Why?**
Signaling is a core feature

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

Design document can be found in this issue: https://github.com/cadence-workflow/cadence-python-client/issues/94 

----
## Summary by Gitar

- **Feature enhancement:**
  - Exposed `retry_policy` support in `StartWorkflowOptions` within `cadence/client.py`.
  - Implemented `_build_start_workflow_request` logic to convert and apply retry policies to workflow requests.

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/cadence/_internal/rpc/error.py
+++ b/cadence/_internal/rpc/error.py
@@ -109,9 +109,7 @@ def map_error(e: AioRpcError) -> error.CadenceRpcError:
             list(not_exists.active_clusters),
         )
     elif details.Is(error_pb2.WorkflowExecutionAlreadyCompletedError.DESCRIPTOR):
-        return error.WorkflowExecutionAlreadyCompletedError(
-            _rpc_details(e), e.code()
-        )
+        return error.WorkflowExecutionAlreadyCompletedError(_rpc_details(e), e.code())
     elif details.Is(error_pb2.DomainNotActiveError.DESCRIPTOR):
         not_active = error_pb2.DomainNotActiveError()
         details.Unpack(not_active)
--- a/cadence/_internal/rpc/error.py
+++ b/cadence/_internal/rpc/error.py
@@ -15,9 +15,9 @@ from cadence.api.v1 import error_pb2
 from cadence import error
 
 
-def _rpc_details(e: AioRpcError) -> str:
+def _rpc_details(e: AioRpcError) -> str | None:
     d = e.details()
-    return d if isinstance(d, str) else ""
+    return d if isinstance(d, str) else None
 
 
 RequestType = TypeVar("RequestType")
--- a/cadence/_internal/rpc/error.py
+++ b/cadence/_internal/rpc/error.py
@@ -15,6 +15,11 @@ from cadence.api.v1 import error_pb2
 from cadence import error
 
 
+def _rpc_details(e: AioRpcError) -> str:
+    d = e.details()
+    return d if isinstance(d, str) else ""
+
+
 RequestType = TypeVar("RequestType")
 ResponseType = TypeVar("ResponseType")
 DoneCallbackType = Callable[[Any], None]
@@ -81,14 +86,14 @@ class CadenceErrorInterceptor(UnaryUnaryClientInterceptor):
 def map_error(e: AioRpcError) -> error.CadenceRpcError:
     status: Status | None = from_call(e)
     if not status or not status.details:
-        return error.CadenceRpcError(e.details(), e.code())
+        return error.CadenceRpcError(_rpc_details(e), e.code())
 
     details = status.details[0]
     if details.Is(error_pb2.WorkflowExecutionAlreadyStartedError.DESCRIPTOR):
         already_started = error_pb2.WorkflowExecutionAlreadyStartedError()
         details.Unpack(already_started)
         return error.WorkflowExecutionAlreadyStartedError(
-            e.details(),
+            _rpc_details(e),
             e.code(),
             already_started.start_request_id,
             already_started.run_id,
@@ -97,19 +102,21 @@ def map_error(e: AioRpcError) -> error.CadenceRpcError:
         not_exists = error_pb2.EntityNotExistsError()
         details.Unpack(not_exists)
         return error.EntityNotExistsError(
-            e.details(),
+            _rpc_details(e),
             e.code(),
             not_exists.current_cluster,
             not_exists.active_cluster,
             list(not_exists.active_clusters),
         )
     elif details.Is(error_pb2.WorkflowExecutionAlreadyCompletedError.DESCRIPTOR):
-        return error.WorkflowExecutionAlreadyCompletedError(e.details(), e.code())
+        return error.WorkflowExecutionAlreadyCompletedError(
+            _rpc_details(e), e.code()
+        )
     elif details.Is(error_pb2.DomainNotActiveError.DESCRIPTOR):
         not_active = error_pb2.DomainNotActiveError()
         details.Unpack(not_active)
         return error.DomainNotActiveError(
-            e.details(),
+            _rpc_details(e),
             e.code(),
             not_active.domain,
             not_active.current_cluster,
@@ -120,7 +127,7 @@ def map_error(e: AioRpcError) -> error.CadenceRpcError:
         not_supported = error_pb2.ClientVersionNotSupportedError()
         details.Unpack(not_supported)
         return error.ClientVersionNotSupportedError(
-            e.details(),
+            _rpc_details(e),
             e.code(),
             not_supported.feature_version,
             not_supported.client_impl,
@@ -130,19 +137,19 @@ def map_error(e: AioRpcError) -> error.CadenceRpcError:
         not_enabled = error_pb2.FeatureNotEnabledError()
         details.Unpack(not_enabled)
         return error.FeatureNotEnabledError(
-            e.details(), e.code(), not_enabled.feature_flag
+            _rpc_details(e), e.code(), not_enabled.feature_flag
         )
     elif details.Is(error_pb2.CancellationAlreadyRequestedError.DESCRIPTOR):
-        return error.CancellationAlreadyRequestedError(e.details(), e.code())
+        return error.CancellationAlreadyRequestedError(_rpc_details(e), e.code())
     elif details.Is(error_pb2.DomainAlreadyExistsError.DESCRIPTOR):
-        return error.DomainAlreadyExistsError(e.details(), e.code())
+        return error.DomainAlreadyExistsError(_rpc_details(e), e.code())
     elif details.Is(error_pb2.LimitExceededError.DESCRIPTOR):
-        return error.LimitExceededError(e.details(), e.code())
+        return error.LimitExceededError(_rpc_details(e), e.code())
     elif details.Is(error_pb2.QueryFailedError.DESCRIPTOR):
-        return error.QueryFailedError(e.details(), e.code())
+        return error.QueryFailedError(_rpc_details(e), e.code())
     elif details.Is(error_pb2.ServiceBusyError.DESCRIPTOR):
         service_busy = error_pb2.ServiceBusyError()
         details.Unpack(service_busy)
-        return error.ServiceBusyError(e.details(), e.code(), service_busy.reason)
+        return error.ServiceBusyError(_rpc_details(e), e.code(), service_busy.reason)
     else:
-        return error.CadenceRpcError(e.details(), e.code())
+        return error.CadenceRpcError(_rpc_details(e), e.code())
--- a/cadence/_internal/rpc/error.py
+++ b/cadence/_internal/rpc/error.py
@@ -15,11 +15,6 @@ from cadence.api.v1 import error_pb2
 from cadence import error
 
 
-def _rpc_details(e: AioRpcError) -> str | None:
-    d = e.details()
-    return d if isinstance(d, str) else None
-
-
 RequestType = TypeVar("RequestType")
 ResponseType = TypeVar("ResponseType")
 DoneCallbackType = Callable[[Any], None]
@@ -86,14 +81,14 @@ class CadenceErrorInterceptor(UnaryUnaryClientInterceptor):
 def map_error(e: AioRpcError) -> error.CadenceRpcError:
     status: Status | None = from_call(e)
     if not status or not status.details:
-        return error.CadenceRpcError(_rpc_details(e), e.code())
+        return error.CadenceRpcError(e.details(), e.code())
 
     details = status.details[0]
     if details.Is(error_pb2.WorkflowExecutionAlreadyStartedError.DESCRIPTOR):
         already_started = error_pb2.WorkflowExecutionAlreadyStartedError()
         details.Unpack(already_started)
         return error.WorkflowExecutionAlreadyStartedError(
-            _rpc_details(e),
+            e.details(),
             e.code(),
             already_started.start_request_id,
             already_started.run_id,
@@ -102,19 +97,19 @@ def map_error(e: AioRpcError) -> error.CadenceRpcError:
         not_exists = error_pb2.EntityNotExistsError()
         details.Unpack(not_exists)
         return error.EntityNotExistsError(
-            _rpc_details(e),
+            e.details(),
             e.code(),
             not_exists.current_cluster,
             not_exists.active_cluster,
             list(not_exists.active_clusters),
         )
     elif details.Is(error_pb2.WorkflowExecutionAlreadyCompletedError.DESCRIPTOR):
-        return error.WorkflowExecutionAlreadyCompletedError(_rpc_details(e), e.code())
+        return error.WorkflowExecutionAlreadyCompletedError(e.details(), e.code())
     elif details.Is(error_pb2.DomainNotActiveError.DESCRIPTOR):
         not_active = error_pb2.DomainNotActiveError()
         details.Unpack(not_active)
         return error.DomainNotActiveError(
-            _rpc_details(e),
+            e.details(),
             e.code(),
             not_active.domain,
             not_active.current_cluster,
@@ -125,7 +120,7 @@ def map_error(e: AioRpcError) -> error.CadenceRpcError:
         not_supported = error_pb2.ClientVersionNotSupportedError()
         details.Unpack(not_supported)
         return error.ClientVersionNotSupportedError(
-            _rpc_details(e),
+            e.details(),
             e.code(),
             not_supported.feature_version,
             not_supported.client_impl,
@@ -135,19 +130,19 @@ def map_error(e: AioRpcError) -> error.CadenceRpcError:
         not_enabled = error_pb2.FeatureNotEnabledError()
         details.Unpack(not_enabled)
         return error.FeatureNotEnabledError(
-            _rpc_details(e), e.code(), not_enabled.feature_flag
+            e.details(), e.code(), not_enabled.feature_flag
         )
     elif details.Is(error_pb2.CancellationAlreadyRequestedError.DESCRIPTOR):
-        return error.CancellationAlreadyRequestedError(_rpc_details(e), e.code())
+        return error.CancellationAlreadyRequestedError(e.details(), e.code())
     elif details.Is(error_pb2.DomainAlreadyExistsError.DESCRIPTOR):
-        return error.DomainAlreadyExistsError(_rpc_details(e), e.code())
+        return error.DomainAlreadyExistsError(e.details(), e.code())
     elif details.Is(error_pb2.LimitExceededError.DESCRIPTOR):
-        return error.LimitExceededError(_rpc_details(e), e.code())
+        return error.LimitExceededError(e.details(), e.code())
     elif details.Is(error_pb2.QueryFailedError.DESCRIPTOR):
-        return error.QueryFailedError(_rpc_details(e), e.code())
+        return error.QueryFailedError(e.details(), e.code())
     elif details.Is(error_pb2.ServiceBusyError.DESCRIPTOR):
         service_busy = error_pb2.ServiceBusyError()
         details.Unpack(service_busy)
-        return error.ServiceBusyError(_rpc_details(e), e.code(), service_busy.reason)
+        return error.ServiceBusyError(e.details(), e.code(), service_busy.reason)
     else:
-        return error.CadenceRpcError(_rpc_details(e), e.code())
+        return error.CadenceRpcError(e.details(), e.code())
--- a/cadence/_internal/rpc/retry.py
+++ b/cadence/_internal/rpc/retry.py
@@ -45,7 +45,6 @@ DEFAULT_RETRY_POLICY = ExponentialRetryPolicy(
     initial_interval=0.02, backoff_coefficient=1.2, max_interval=6, max_attempts=0
 )
 GET_WORKFLOW_HISTORY = b"/uber.cadence.api.v1.WorkflowAPI/GetWorkflowExecutionHistory"
-GET_WORKFLOW_HISTORY_STR = GET_WORKFLOW_HISTORY.decode("ascii")
 
 
 class RetryInterceptor(UnaryUnaryClientInterceptor):
@@ -100,10 +99,10 @@ class RetryInterceptor(UnaryUnaryClientInterceptor):
 
 def is_retryable(err: CadenceRpcError, call_details: ClientCallDetails) -> bool:
     # Handle requests to the passive side, matching the Go and Java Clients
-    if call_details.method in (
-        GET_WORKFLOW_HISTORY,
-        GET_WORKFLOW_HISTORY_STR,
-    ) and isinstance(err, EntityNotExistsError):
+    if (
+        call_details.method == GET_WORKFLOW_HISTORY  # type: ignore[comparison-overlap]
+        and isinstance(err, EntityNotExistsError)
+    ):
         return (
             err.active_cluster is not None
             and err.current_cluster is not None
--- a/cadence/_internal/rpc/retry.py
+++ b/cadence/_internal/rpc/retry.py
@@ -45,6 +45,7 @@ DEFAULT_RETRY_POLICY = ExponentialRetryPolicy(
     initial_interval=0.02, backoff_coefficient=1.2, max_interval=6, max_attempts=0
 )
 GET_WORKFLOW_HISTORY = b"/uber.cadence.api.v1.WorkflowAPI/GetWorkflowExecutionHistory"
+GET_WORKFLOW_HISTORY_STR = GET_WORKFLOW_HISTORY.decode("ascii")
 
 
 class RetryInterceptor(UnaryUnaryClientInterceptor):
@@ -60,6 +61,8 @@ class RetryInterceptor(UnaryUnaryClientInterceptor):
     ) -> Any:
         loop = asyncio.get_running_loop()
         expiration_interval = client_call_details.timeout
+        if expiration_interval is None:
+            expiration_interval = float("inf")
         start_time = loop.time()
         deadline = start_time + expiration_interval
 
@@ -68,7 +71,9 @@ class RetryInterceptor(UnaryUnaryClientInterceptor):
             remaining = deadline - loop.time()
             # Namedtuple methods start with an underscore to avoid conflicts and aren't actually private
             # noinspection PyProtectedMember
-            call_details = client_call_details._replace(timeout=remaining)
+            call_details = client_call_details._replace(  # type: ignore[attr-defined]
+                timeout=remaining
+            )
             rpc_call = await continuation(call_details, request)
             try:
                 await rpc_call
@@ -95,7 +100,10 @@ class RetryInterceptor(UnaryUnaryClientInterceptor):
 
 def is_retryable(err: CadenceRpcError, call_details: ClientCallDetails) -> bool:
     # Handle requests to the passive side, matching the Go and Java Clients
-    if call_details.method == GET_WORKFLOW_HISTORY and isinstance(
+    if call_details.method in (
+        GET_WORKFLOW_HISTORY,
+        GET_WORKFLOW_HISTORY_STR,
+    ) and isinstance(
         err, EntityNotExistsError
     ):
         return (
--- a/cadence/_internal/rpc/retry.py
+++ b/cadence/_internal/rpc/retry.py
@@ -103,9 +103,7 @@ def is_retryable(err: CadenceRpcError, call_details: ClientCallDetails) -> bool:
     if call_details.method in (
         GET_WORKFLOW_HISTORY,
         GET_WORKFLOW_HISTORY_STR,
-    ) and isinstance(
-        err, EntityNotExistsError
-    ):
+    ) and isinstance(err, EntityNotExistsError):
         return (
             err.active_cluster is not None
             and err.current_cluster is not None
--- a/cadence/_internal/rpc/retry.py
+++ b/cadence/_internal/rpc/retry.py
@@ -71,8 +71,11 @@ class RetryInterceptor(UnaryUnaryClientInterceptor):
             call_details = client_call_details._replace(timeout=remaining)
             rpc_call = await continuation(call_details, request)
             try:
-                # Return the result directly if success. GRPC will wrap it back into a UnaryUnaryCall
-                return await rpc_call
+                await rpc_call
+                # Return the call object (not the raw response) so outer interceptors
+                # that rely on UnaryUnaryCall methods like add_done_callback still work
+                # (e.g. opentelemetry-instrumentation-grpc).
+                return rpc_call
             except CadenceRpcError as e:
                 err = e
 
--- a/cadence/_internal/workflow/context.py
+++ b/cadence/_internal/workflow/context.py
@@ -31,7 +31,6 @@ class Context(WorkflowContext):
         self._replay_mode = True
         self._replay_current_time_milliseconds: Optional[int] = None
         self._decision_manager = decision_manager
-        self._waiters: list[tuple[Callable[[], bool], Any]] = []
 
     def info(self) -> WorkflowInfo:
         return self._info
@@ -94,7 +93,6 @@ class Context(WorkflowContext):
         )
 
         future = self._decision_manager.schedule_activity(schedule_attributes)
-        future.add_done_callback(lambda _: self.notify_state_changed())
         result_payload = await future
 
         result = self.data_converter().from_data(result_payload, [result_type])[0]
@@ -109,7 +107,6 @@ class Context(WorkflowContext):
                 start_to_fire_timeout=duration,
             )
         )
-        future.add_done_callback(lambda _: self.notify_state_changed())
         await future
 
     def set_replay_mode(self, replay: bool) -> None:
@@ -129,35 +126,7 @@ class Context(WorkflowContext):
         return self._replay_current_time_milliseconds
 
     async def wait_condition(self, predicate: Callable[[], bool]) -> None:
-        if predicate():
-            return
-        future = self._decision_manager._event_loop.create_future()
-        self._waiters.append((predicate, future))
-        await future
-
-    def notify_state_changed(self) -> None:
-        """Re-evaluate all wait_condition predicates. Resolve those that are now True.
-
-        If a predicate raises, the exception is set on that specific waiter's
-        future so the ``await wait_condition(...)`` call receives a
-        deterministic error.  Other waiters continue to be evaluated
-        normally — one broken predicate does not block or deadlock
-        unrelated waits.
-        """
-        remaining: list[tuple[Callable[[], bool], Any]] = []
-        for predicate, future in self._waiters:
-            if future.done():
-                continue
-            try:
-                result = predicate()
-            except Exception as exc:
-                future.set_exception(exc)
-                continue
-            if result:
-                future.set_result(None)
-            else:
-                remaining.append((predicate, future))
-        self._waiters = remaining
+        await self._decision_manager.create_waiter(predicate)
 
     @contextmanager
     def _activate(self) -> Iterator["Context"]:
--- a/cadence/_internal/workflow/context.py
+++ b/cadence/_internal/workflow/context.py
@@ -136,12 +136,24 @@ class Context(WorkflowContext):
         await future
 
     def notify_state_changed(self) -> None:
-        """Re-evaluate all wait_condition predicates. Resolve those that are now True."""
+        """Re-evaluate all wait_condition predicates. Resolve those that are now True.
+
+        If a predicate raises, the exception is set on that specific waiter's
+        future so the ``await wait_condition(...)`` call receives a
+        deterministic error.  Other waiters continue to be evaluated
+        normally — one broken predicate does not block or deadlock
+        unrelated waits.
+        """
         remaining: list[tuple[Callable[[], bool], Any]] = []
         for predicate, future in self._waiters:
             if future.done():
                 continue
-            if predicate():
+            try:
+                result = predicate()
+            except Exception as exc:
+                future.set_exception(exc)
+                continue
+            if result:
                 future.set_result(None)
             else:
                 remaining.append((predicate, future))
--- a/cadence/_internal/workflow/context.py
+++ b/cadence/_internal/workflow/context.py
@@ -1,4 +1,5 @@
 from contextlib import contextmanager
+from asyncio import get_running_loop
 from datetime import timedelta
 from math import ceil
 from typing import Iterator, Optional, Any, Unpack, Type, cast, Callable
@@ -31,13 +32,11 @@ class Context(WorkflowContext):
         self,
         info: WorkflowInfo,
         decision_manager: DecisionManager,
-        event_loop: DeterministicEventLoop,
     ):
         self._info = info
         self._replay_mode = True
         self._replay_current_time_milliseconds: Optional[int] = None
         self._decision_manager = decision_manager
-        self._event_loop = event_loop
 
     def info(self) -> WorkflowInfo:
         return self._info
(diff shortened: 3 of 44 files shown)


timl3136:
Gitar fix CI

timl3136:
Gitar fix CI

cadence/_internal/rpc/error.py
```diff
@@ -15,6 +15,11 @@
 from cadence import error
 
 
+def _rpc_details(e: AioRpcError) -> str:
+    d = e.details()
+    return d if isinstance(d, str) else ""
```
shijiesheng: There is no difference of "None" and "". If we really want to do it right, it should omit message argument in 
```
class CadenceRpcError(Exception):
    def __init__(self, message: str, code: grpc.StatusCode, *args):
        super().__init__(message, code, *args) # We should omit message if this is None. 
        self.code = code
```

cadence/_internal/rpc/retry.py
```diff
@@ -45,6 +45,7 @@ def next_delay(
     initial_interval=0.02, backoff_coefficient=1.2, max_interval=6, max_attempts=0
 )
 GET_WORKFLOW_HISTORY = b"/uber.cadence.api.v1.WorkflowAPI/GetWorkflowExecutionHistory"
+GET_WORKFLOW_HISTORY_STR = GET_WORKFLOW_HISTORY.decode("ascii")
```
shijiesheng: revert the change and keep the raw byte and add type ignore. This is a bug but was fixed only in new versions of grpcio https://github.com/grpc/grpc/pull/40215

cadence/_internal/rpc/retry.py
```diff
@@ -68,7 +71,9 @@ async def intercept_unary_unary(
             remaining = deadline - loop.time()
             # Namedtuple methods start with an underscore to avoid conflicts and aren't actually private
             # noinspection PyProtectedMember
-            call_details = client_call_details._replace(timeout=remaining)
+            call_details = client_call_details._replace(  # type: ignore[attr-defined]
```
shijiesheng, line 73: just a comment that python introduced copy.replace in 3.13 for this. But I guess we can still rely on this for now.

cadence/_internal/workflow/context.py
```diff
@@ -125,11 +128,44 @@ def get_replay_current_time_milliseconds(self) -> Optional[int]:
...
+
     @contextmanager
     def _activate(self) -> Iterator["Context"]:
         token = WorkflowContext._var.set(self)
-        yield self
-        WorkflowContext._var.reset(token)
+        try:
+            yield self
+        finally:
+            WorkflowContext._var.reset(token)
```
shijiesheng, line 144: Why do we need this try?

cadence/_internal/rpc/error.py
timl3136: Thanks

cadence/_internal/workflow/context.py
```diff
@@ -103,11 +104,13 @@ async def execute_activity(
     async def start_timer(self, duration: timedelta):
         if duration.total_seconds() <= 0:  # shortcut
             return
-        await self._decision_manager.start_timer(
+        future = self._decision_manager.start_timer(
             StartTimerDecisionAttributes(
                 start_to_fire_timeout=duration,
             )
         )
+        future.add_done_callback(lambda _: self.notify_state_changed())
```
shijiesheng: We no longer need this extra trigger and can do waiters check in _run_once(). It's more intuitive to refresh waiters check every run_once cycle. You can consider run_once as run all these coroutines "at the same time" (but in fact they are run in FIFO order)

cadence/_internal/workflow/context.py
```diff
@@ -125,11 +128,44 @@ def get_replay_current_time_milliseconds(self) -> Optional[int]:
         """Get the current replay time in milliseconds."""
         return self._replay_current_time_milliseconds
 
+    async def wait_condition(self, predicate: Callable[[], bool]) -> None:
+        if predicate():
+            return
+        future = self._decision_manager._event_loop.create_future()
+        self._waiters.append((predicate, future))
+        await future
```
shijiesheng: Hide the implementation and manage waiters inside eventloop. Specifically, 

1. add Waiter entity (extended from awaitable) 
2. add create_waiter() to DeterministicEventLoop
```suggestion
        await self._decision_manager._event_loop.create_waiter(predicate)
```

cadence/_internal/workflow/context.py
```diff
@@ -125,11 +128,44 @@ def get_replay_current_time_milliseconds(self) -> Optional[int]:
...
+
+    def notify_state_changed(self) -> None:
```
shijiesheng: not needed as discussed above

cadence/_internal/rpc/error.py
```diff
@@ -15,6 +15,11 @@
 from cadence import error
 
 
+def _rpc_details(e: AioRpcError) -> str | None:
```
shijiesheng: it's unnecessary now because details() already returns "str | None"

tests/integration_tests/workflow/test_signals.py
```diff
@@ -0,0 +1,179 @@
...
+from cadence.api.v1.service_workflow_pb2 import (
+    GetWorkflowExecutionHistoryRequest,
+    GetWorkflowExecutionHistoryResponse,
+)
+from tests.integration_tests.helper import CadenceHelper, DOMAIN_NAME
+
+registry = Registry()
+
+
+class _SignalWaitBase:
```
shijiesheng, line 15: Could you add a test that demonstrate that Cadence Python SDK can address this edge case? https://github.com/temporalio/sdk-python/issues/618

cadence/_internal/workflow/context.py
```diff
@@ -92,9 +92,8 @@ async def execute_activity(
             request_local_dispatch=False,
         )
 
-        result_payload = await self._decision_manager.schedule_activity(
-            schedule_attributes
-        )
+        future = self._decision_manager.schedule_activity(schedule_attributes)
+        result_payload = await future
```
shijiesheng, line 102: nit: not necessary (maybe easier to put a breakpoint) but this comment can be ignored

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -20,6 +20,43 @@ def __init__(self, *args) -> None:
 _T = TypeVar("_T")
 
 
+class Waiter:
```
shijiesheng:
```suggestion
class Waiter(Awaitable[None]):
```

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -20,6 +20,43 @@ def __init__(self, *args) -> None:
...
+    def done(self) -> bool:
+        return self._future.done()
+
+    def exception(self) -> BaseException | None:
+        return self._future.exception()
+
+    def result(self) -> None:
+        self._future.result()
+
+    def _poll(self) -> bool:
```
shijiesheng: nit: make it public rather than protected

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -20,6 +20,43 @@ def __init__(self, *args) -> None:
...
+            return True
+        try:
+            ready = self._predicate()
+        except BaseException as exc:
+            self._future.set_exception(exc)
+            return True
+        if ready:
+            self._future.set_result(None)
+            return True
+        return False
```
shijiesheng: nit
```suggestion
        try:
            ready = self._predicate()
            if ready:
                self._future.set_result(None)
                return True
            return False
        except BaseException as exc:
            self._future.set_exception(exc)
            return True
```

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -141,13 +179,45 @@ def create_task(
...
         ntodo = len(self._ready)
         for i in range(ntodo):
             handle = self._ready.popleft()
             if handle._cancelled:
                 continue
-            handle._run()
+            self._run_handle(handle)
+
+        if self._waiters:
+            self._waiters = [w for w in self._waiters if not w._poll()]
```
shijiesheng: implementation here isn't right. We should instead break for-loop immediately on any True poll here. Because the future callback could have impact on other waiters. 

The correct logic is below:

_run_once():
  ...
  for w in self._waiters:
    if self._waiters.poll():
       self._waiters.pop(w) # if you use OrderedDict
       break

run_until_yield():
  # likely no change is needed because poll == True will set Future and that will add a task back to ready
  # but maybe just add a unit test to ensure that's the case

pyproject.toml
```diff
@@ -96,6 +97,8 @@ exclude = [
     "cadence/api",
     "cadence/api/.*",
     "cadence/sample",
+    "cadence/contrib/openai",
```
shijiesheng: There is another PR for this already. Do you still want to land that change first?

cadence/_internal/workflow/deterministic_event_loop.py
timl3136: I don't think the return false between try and except here is correct.

pyproject.toml
timl3136: I will move this to another pr

cadence/_internal/workflow/deterministic_event_loop.py
shijiesheng: I submitted with wrong indentation. It should be correct now.

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -141,13 +179,47 @@ def create_task(
...
+        except (SystemExit, KeyboardInterrupt):
+            raise
+        except BaseException as exc:
+            handle._loop.call_exception_handler(  # type: ignore[attr-defined]
+                {
+                    "message": "Exception in callback %r" % handle,
+                    "exception": exc,
+                    "handle": handle,
+                }
+            )
```
shijiesheng: This suggests issues in Waiter implementation and should be addressed there

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -162,13 +162,23 @@ def _process_decision_events(
...
+                # Phase 2: Apply input events in history order.
                 for event in decision_events.input:
-                    self._apply_input_event(event)
+                    if (
+                        event.WhichOneof("attributes")
+                        == "workflow_execution_signaled_event_attributes"
+                    ):
+                        self._workflow_instance.handle_signal_event(event)
+                    else:
+                        self._apply_input_event(event)
```
shijiesheng: put logic in self._apply_input_event() instead

cadence/workflow.py
```diff
@@ -260,11 +270,31 @@ def signal(name: str | None = None) -> Callable[[T], T]:
...
 
-    Example:
+    Signal handlers must be **synchronous** functions that mutate workflow
+    state and return ``None``.  They run on the deterministic event loop
+    in the same thread as the workflow; keep them fast and free of I/O.
+
+    Async signal handlers are not supported and will be rejected at
+    definition time.  The SDK's deterministic event loop does not track
+    detached tasks, so an ``async def`` handler could silently diverge
+    from the replay-safe execution model.
```
shijiesheng: I'm pretty sure we can support async

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -141,13 +179,47 @@ def create_task(
...
-            handle._run()
+            self._run_handle(handle)
+
+        for i, w in enumerate(self._waiters):
+            if w.poll():
+                del self._waiters[i]
+                break
+
+    @staticmethod
+    def _run_handle(handle: events.Handle) -> None:
```
timl3136: not necessary, handle.run would work here

cadence/_internal/workflow/statemachine/decision_manager.py
timl3136, line 69: add signal decision task here

cadence/signal.py
```diff
@@ -75,16 +79,40 @@ def params(self) -> list[SignalParameter]:
...
     @property
     def wrapped(self) -> Callable[P, T]:
         """Get the wrapped signal handler function."""
         return self._wrapped
 
+    def params_from_payload(self, data_converter: Any, payload: Payload) -> list[Any]:
+        if not self._params:
+            return []
+
+        payload_count = _count_payload_values(payload)
```
natemort: Payloads are an opaque blob. They might not contain JSON at all, it's entirely up to the DataConverter to make sense of what's in there. Preprocessing it isn't correct.

If we need to unify the DataConverter's output with the functions signature then we need to do that everywhere (Activities, Workflows), and so we should update FnSignature so that we only need to handle that once.

cadence/workflow.py
```diff
@@ -270,11 +280,31 @@ def signal(name: str | None = None) -> Callable[[T], T]:
...
+        def approve(self, approved: bool) -> None:
             self.approved = approved
 
+        @workflow.signal(name="async_approval")
+        async def approve_async(self, approved: bool) -> None:
+            self.approved = approved
+            await workflow.execute_activity("notify", ...)
+
+    Concurrency constraints:
+        * Do **not** use native threads or real-loop ``asyncio`` primitives
```
natemort: Using those should be fine, why wouldn't they be replay safe?

At a glance, both look deterministic.

cadence/signal.py
```diff
@@ -106,11 +134,10 @@ def wrap(
...
         name = opts.get("name") or fn.__qualname__
-        is_async = inspect.iscoroutinefunction(fn)
         params = _get_signal_signature(fn)
         _validate_signal_return_type(fn)
 
-        return SignalDefinition[P, T](fn, name, params, is_async)
+        return SignalDefinition[P, T](fn, name, params)
 
 
 def _validate_signal_return_type(fn: Callable) -> None:
```
natemort: Can we update FnSignature to handle this?

cadence/signal.py
natemort: Let's reuse the types in FnSignature rather than duplicating it

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -24,6 +27,9 @@ def __init__(
         self._data_converter = data_converter
         self._instance = workflow_definition.cls()  # construct a new workflow object
         self._task: Optional[Task[Payload]] = None
+        # Strong references to in-flight async signal handler tasks.
+        self._signal_tasks: set[Task[Any]] = set()
+        self._signal_error: Optional[Exception] = None
```
natemort: Nit: I wonder if we can handle this more generally in DeterministicEventLoop. We want this behavior for any unhandled exception, whether it's from a signal or the user is submitting their own tasks to the event loop.

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -238,17 +242,16 @@ def _apply_input_event(self, event: history.HistoryEvent) -> None:
                 "replay_mode": self._context.is_replay_mode(),
             },
         )
-        # start workflow on workflow started event
-        if (
-            event.WhichOneof("attributes")
-            == "workflow_execution_started_event_attributes"
-        ):
+        attr = event.WhichOneof("attributes")
+        if attr == "workflow_execution_signaled_event_attributes":
```
natemort: Nit: We could use singledispatch to call different methods based on the type of the attributes rather than using these conditions. If we don't have a specific handler in WorkflowEngine then our fallback is the DecisionManager.

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +57,65 @@ def get_result(self) -> Optional[Payload]:
...
+            args = signal_def.params_from_payload(self._data_converter, payload)
+        except Exception as e:
+            logger.warning(
+                "Failed to decode payload for signal '%s', dropping: %s",
+                signal_name,
+                e,
+            )
+            return
+
+        handler = signal_def.wrapped.__get__(self._instance)
```
natemort: Can this be simplified to the following?
```
result = signal_def(self._instance, *args)
```

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -20,6 +20,41 @@ def __init__(self, *args) -> None:
 _T = TypeVar("_T")
 
 
+class Waiter(Awaitable[None]):
```
natemort: Nit: We could put this in its own file and test it independently.

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -149,6 +192,11 @@ def _run_once(self) -> None:
                 continue
             handle._run()
 
+        for i, w in enumerate(self._waiters):
+            if w.poll():
```
natemort, line 168: I think this is subtly incorrect. If I create two Waiters and only wait on one of them, we're going to prematurely break out of the event loop:
```
workflow.wait_condition(lambda: return ...)
await workflow.wait_condition(lambda: return ...)
```
Execution will:
1. Add the above workflow function to _ready
2. Run it and remove the function from _ready, which will add these two waiters
3. Run the condition in the first waiter, which will then resolve and break out of the loop. _run_once will return.

It's hard to provide an exact code snippet while writing this because we short-circuit any Waiters that are immediately true, but we could create one by having the lambdas check if some future is completed and then asynchronously completing them.

We should instead resume running the workflow function

This logic is assuming that resolving the Waiter is going to add a handle to _ready, which might not be true.

We only want to break out of the _waiters loop if new handles were added to _ready.

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -32,8 +67,12 @@ def __init__(self) -> None:
         self._thread_id: int | None = None  # indicate if the event loop is running
         self._debug: bool = False
         self._ready: collections.deque[events.Handle] = collections.deque()
+        self._waiters: list[Waiter] = []
         self._stopping: bool = False
         self._closed: bool = False
+        # First user callback / task failure this tick (``Exception`` only;
+        # ``SystemExit`` / ``KeyboardInterrupt`` are not captured).
+        self._first_exception: Exception | None = None
```
shijiesheng: this looks suspicious. Why is it needed?

cadence/_internal/workflow/deterministic_event_loop.py
shijiesheng: Good catch! (It's possible that a misuse would cause no callbacks). 

Solution 1: 
keep _run_once logic as it is. In run_until_yield, instead of checking _ready queue is empty, check _ready queue empty AND all waiters' predicates are false. This will create more unnecessary iterations on predicates check because if we known there are no additional callback, it SHOULD stay the same. But this is very intuitive and easy to maintain.

Solution 2:
As Nate suggested, poll true doesn't break unless it has _callbacks. Waiter should extend Future and thus you can just access it through w._callbacks(). This is minimal change but need additional comments for explanation.

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -24,6 +27,8 @@ def __init__(
         self._data_converter = data_converter
         self._instance = workflow_definition.cls()  # construct a new workflow object
         self._task: Optional[Task[Payload]] = None
+        # Strong references to in-flight async signal handler tasks.
+        self._signal_tasks: set[Task[Any]] = set()
```
shijiesheng, line 31: Instead of adding exceptions storage in eventloop, you access these coroutines here.

cadence/_internal/workflow/deterministic_event_loop.py
shijiesheng: You can access signal coroutines when they are created rather than access them from here.

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -461,14 +492,23 @@ def default_exception_handler(self, context: dict[str, Any]) -> None:
...
         # Long term, we need some combination of failing decision tasks or workflows based on these errors.
+        exception = context.get("exception")
+        if isinstance(exception, Exception) and self._first_exception is None:
+            self._first_exception = exception
+
         message = context.get("message")
         if not message:
             message = "Unhandled exception in event loop"
 
-        exception = context.get("exception")
```
shijiesheng: as mentioned above, not needed here. Instead, move the logic out

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -46,6 +46,14 @@ def __init__(self, info: WorkflowInfo, workflow_definition: WorkflowDefinition):
...
         )
         self._context = Context(info, self._decision_manager)
+        self._input_event_handlers: dict[str, Callable[[HistoryEvent], None]] = {
+            "workflow_execution_signaled_event_attributes": (
+                self._handle_signaled_input_event
+            ),
+            "workflow_execution_started_event_attributes": (
+                self._handle_started_input_event
+            ),
+        }
```
shijiesheng: Just a comment, we should consolidate the two: input_event_handlers and eventdispatcher

cadence/_internal/workflow/deterministic_event_loop.py
timl3136: You are right here, this is only needed if we are using stock eventloop. I will remove this from the code

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -238,19 +253,28 @@ def _apply_input_event(self, event: history.HistoryEvent) -> None:
...
-            == "workflow_execution_started_event_attributes"
-        ):
-            started_attrs: WorkflowExecutionStartedEventAttributes = (
-                event.workflow_execution_started_event_attributes
-            )
-            if started_attrs and hasattr(started_attrs, "input"):
-                self._workflow_instance.start(started_attrs.input)
-
+        attr = event.WhichOneof("attributes")
+        event_attributes = getattr(event, attr)
```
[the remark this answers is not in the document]
timl3136: gitar fix

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -238,19 +253,28 @@ def _apply_input_event(self, event: history.HistoryEvent) -> None:
...
+        action = input_events.handlers.get(type(event_attributes))
+        if action is not None:
+            action.fn(self, event_attributes)
+            if isinstance(event_attributes, WorkflowExecutionSignaledEventAttributes):
+                return
```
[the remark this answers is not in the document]
timl3136: gitar fix

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -462,13 +490,12 @@ def default_exception_handler(self, context: dict[str, Any]) -> None:
...
     def call_exception_handler(self, context: dict[str, Any]) -> None:
-        # This is called if a task has an unhandled exception. Short term, it's helpful to log these for debugging.
-        # Long term, we need some combination of failing decision tasks or workflows based on these errors.
+        exception = context.get("exception")
+
         message = context.get("message")
         if not message:
             message = "Unhandled exception in event loop"
 
-        exception = context.get("exception")
```
shijiesheng: unnecessary changes. This is copied from BaseEventloop implementation, and we want to keep them the same as much as possible

cadence/_internal/workflow/deterministic_event_loop.py
```diff
@@ -493,6 +520,9 @@ def call_exception_handler(self, context: dict[str, Any]) -> None:
 
         logger.error("\n".join(log_lines), exc_info=exc_info)
 
+        if isinstance(exception, Exception):
+            raise exception
+
```
shijiesheng: don't raise it. You get the exception from the created tasks in signal handling.

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -23,12 +24,14 @@
...
     HistoryEvent,
     WorkflowExecutionStartedEventAttributes,
+    WorkflowExecutionSignaledEventAttributes,
 )
 from cadence.api.v1.tasklist_pb2 import TaskList
 from cadence.error import ContinueAsNewError
 from cadence.workflow import WorkflowDefinition, WorkflowInfo
 
 logger = logging.getLogger(__name__)
+input_events = EventDispatcher()
```
shijiesheng: move it to statemachine folder and decision_manager can do the handling there.

cadence/_internal/workflow/workflow_engine.py
shijiesheng: code duplication by AI again.

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +59,49 @@ def get_result(self) -> Optional[Payload]:
...
+            self._signal_tasks.add(task)
+            task.add_done_callback(self._on_signal_task_done)
+
+    def _on_signal_task_done(self, task: Task[Any]) -> None:
+        self._signal_tasks.discard(task)
+        if task.cancelled():
+            return
+        exc = task.exception()
+        if isinstance(exc, Exception):
+            raise exc
```
shijiesheng: would be much simpler if you just create_task here. Signal Method should be async and you can wrap sync with async

cadence/_internal/workflow/deterministic_event_loop.py
timl3136: Thanks, I have addressed this issue.

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -119,6 +151,14 @@ def complete_workflow(self, decision: decision.Decision) -> None:
 
         self._add_state_machine(CompletionStateMachine(decision))
 
+    def create_waiter(self, predicate: Callable[[], bool]) -> Waiter:
```
natemort: Nit: I don't think this benefits from being in the DecisionManager, it's independent of the decision state machines or any of their tracking.

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -137,6 +177,21 @@ def _add_state_machine(self, state: DecisionStateMachine) -> None:
         self.state_machines[decision_id] = state
         self.aliases[(decision_id.decision_type, decision_id.id)] = state
 
+    # ----- Input event routing -----
+
+    def handle_input_event(
```
natemort: Nit: Sorry, I don't think my previous comment was super clear.

I was suggesting that we use [single dispatch method](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod) to automatically route to a specific function in WorkflowEngine. The fallback scenario should be to pass it to DecisionManager, but for signals and StartWorkflow we should handle it there.

I don't think this logic belongs in DecisionManager as its unrelated to the DecisionStateMachines.

cadence/signal.py
```diff
@@ -71,20 +62,25 @@ def name(self) -> str:
         return self._name
 
     @property
-    def params(self) -> list[SignalParameter]:
-        """Get the signal parameters."""
-        return self._params
+    def signature(self) -> FnSignature:
```
natemort: Nit: Minimize exposing properties, as this is part of our public API. I don't think `signature` or `params` are used by the implementation, and even then I'd probably prefer we access a "private" method/field over making this part of our API. We don't want users to rely on these.

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -29,9 +34,36 @@
...
+
+input_events: EventDispatcher = EventDispatcher()
+
+
+@input_events.event()
+def _handle_signaled_input_event(
+    workflow_instance: WorkflowInstance,
+    attrs: WorkflowExecutionSignaledEventAttributes,
+) -> None:
+    workflow_instance.handle_signal_attributes(attrs)
```
shijiesheng:
```suggestion
    workflow_instance.handle_signal(attrs.signal_name, attrs.input)
```

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +64,46 @@ def get_result(self) -> Optional[Payload]:
         if self._task is None or not self._task.done():
             return None
         return self._task.result()
+
+    def handle_signal_attributes(
+        self, attrs: WorkflowExecutionSignaledEventAttributes
```
shijiesheng:
```suggestion
    def handle_signal(
        self, signal_name: str, payload: Payload
```

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +64,46 @@ def get_result(self) -> Optional[Payload]:
...
+    ) -> None:
+        self._loop.call_soon(self._invoke_signal, attrs.signal_name, attrs.input)
+
+    def _invoke_signal(self, signal_name: str, payload: Payload) -> None:
```
shijiesheng:
```suggestion

```

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +64,46 @@ def get_result(self) -> Optional[Payload]:
...
+            logger.warning(
+                "Failed to decode payload for signal '%s', dropping: %s",
+                signal_name,
+                e,
+            )
+            return
+
+        task = self._loop.create_task(self._run_signal(signal_def, args))
+        self._signal_tasks.add(task)
+        task.add_done_callback(self._on_signal_task_done)
```
shijiesheng:
```suggestion
        self._signal_tasks.append((signal_name, task))

    def get_signals(self) -> list[tuple[str, Task[Any]]]:
        return self._signal_tasks
```

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -1,12 +1,18 @@
+import inspect
 import logging
 from asyncio import Task
 from typing import Any, Optional, Callable, Awaitable
```
shijiesheng, line 4:
```suggestion
from typing import Any, Optional, Callable, Awaitable, OrderedDict
```

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +64,46 @@ def get_result(self) -> Optional[Payload]:
...
+        if inspect.iscoroutine(result):
+            await result
+
+    def _on_signal_task_done(self, task: Task[Any]) -> None:
+        self._signal_tasks.discard(task)
+        if task.cancelled():
+            return
+        exc = task.exception()
+        if isinstance(exc, Exception) and self._signal_failure is None:
+            self._signal_failure = exc
```
shijiesheng:
```suggestion

```

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -24,6 +30,13 @@ def __init__(
         self._data_converter = data_converter
         self._instance = workflow_definition.cls()  # construct a new workflow object
         self._task: Optional[Task[Payload]] = None
+        # Strong references to in-flight async signal handler tasks.
+        self._signal_tasks: set[Task[Any]] = set()
+        # fail workflow execution if a signal handler raises an exception
+        self._signal_failure: Optional[Exception] = None
+
+    def get_signal_failure(self) -> Optional[Exception]:
+        return self._signal_failure
```
shijiesheng, line 36:
```suggestion
        self._signal_tasks: list[tuple(str,Task[Any])] = []
```

tests/cadence/_internal/workflow/test_signal_handling.py
```diff
@@ -0,0 +1,1390 @@
...
+    DecisionTaskCompletedEventAttributes,
+    DecisionTaskScheduledEventAttributes,
+    DecisionTaskStartedEventAttributes,
+    HistoryEvent,
+    WorkflowExecutionStartedEventAttributes,
+    WorkflowExecutionSignaledEventAttributes,
+)
+from cadence._internal.workflow.workflow_engine import WorkflowEngine
+from cadence import workflow
+from cadence.data_converter import DefaultDataConverter
```
shijiesheng, line 22:
```suggestion
from cadence.data_converter import DefaultDataConverter
from cadence.error import SignalFailure
```

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -12,17 +12,15 @@
...
     CompleteWorkflowExecutionDecisionAttributes,
     ContinueAsNewWorkflowExecutionDecisionAttributes,
 )
+from cadence.api.v1.common_pb2 import Failure, WorkflowType
 from cadence.api.v1.history_pb2 import (
     HistoryEvent,
-    WorkflowExecutionStartedEventAttributes,
 )
 from cadence.api.v1.tasklist_pb2 import TaskList
 from cadence.error import ContinueAsNewError
```
shijiesheng, line 29:
```suggestion
from cadence.error import ContinueAsNewError, SignalFailure
```

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +64,46 @@ def get_result(self) -> Optional[Payload]:
...
+        try:
+            args = signal_def.params_from_payload(self._data_converter, payload)
+        except Exception as e:
+            logger.warning(
+                "Failed to decode payload for signal '%s', dropping: %s",
+                signal_name,
+                e,
+            )
+            return
+
```
shijiesheng, line 83:
```suggestion

        # TODO consider removing signal tasks after they are done. Keep them around for debugging.
```

tests/cadence/_internal/workflow/test_signal_handling.py
```diff
@@ -0,0 +1,1390 @@
...
+            HistoryEvent(
+                event_id=4,
+                decision_task_started_event_attributes=DecisionTaskStartedEventAttributes(
+                    scheduled_event_id=3,
+                ),
+            ),
+        ]
+
+        engine.process_decision(events)
+        assert engine._workflow_instance._signal_tasks == set()
```
shijiesheng, line 1318:
```suggestion
        assert len(engine._workflow_instance._signal_tasks) == 0
```

tests/cadence/_internal/workflow/test_signal_handling.py
```diff
@@ -0,0 +1,1390 @@
...
+                    scheduled_event_id=3,
+                ),
+            ),
+        ]
+
+        result = engine.process_decision(events)
+        assert len(result.decisions) == 1
+        fail_attrs = result.decisions[0].fail_workflow_execution_decision_attributes
+        assert fail_attrs is not None
+        assert "handler failed on: boom" in fail_attrs.failure.details.decode()
```
shijiesheng:
```suggestion
        with pytest.raises(SignalFailure) as excinfo:
            engine.process_decision(events)
        assert "handler failed on: boom" in str(excinfo.value)
```

tests/cadence/_internal/workflow/test_signal_handling.py
```diff
@@ -0,0 +1,1390 @@
...
+        assert engine.is_done()
+        assert len(result.decisions) == 1
+        assert (
+            "Received signal 'nonexistent_signal' but no handler registered"
+            in caplog.text
+        )
+        completion = result.decisions[0].complete_workflow_execution_decision_attributes
+        assert DATA_CONVERTER.from_data(completion.result, [str]) == ["notified"]
+
+    def test_signal_handler_exception_fails_workflow(self):
```
shijiesheng:
```suggestion
    def test_signal_handler_exception_fails_decision(self):
```

tests/cadence/_internal/workflow/test_signal_handling.py
```diff
@@ -0,0 +1,1390 @@
...
+                    scheduled_event_id=4,
+                ),
+            ),
+        ]
+
+        result = engine.process_decision(events)
+        assert engine.is_done()
+        completion = result.decisions[0].complete_workflow_execution_decision_attributes
+        assert DATA_CONVERTER.from_data(completion.result, [str]) == ["sync:a,async:b"]
+
```
shijiesheng, line 1293:
```suggestion

    @pytest.mark.skip(reason="TODO: keep signal tasks around for debugging")
```

tests/cadence/_internal/workflow/test_signal_handling.py
```diff
@@ -0,0 +1,1390 @@
...
+    DecisionTaskStartedEventAttributes,
+    HistoryEvent,
+    WorkflowExecutionStartedEventAttributes,
+    WorkflowExecutionSignaledEventAttributes,
+)
+from cadence._internal.workflow.workflow_engine import WorkflowEngine
+from cadence import workflow
+from cadence.data_converter import DefaultDataConverter
+from cadence.workflow import WorkflowInfo, WorkflowDefinition, WorkflowDefinitionOptions
+
```
shijiesheng, line 25:
```suggestion
import pytest
```

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -162,15 +160,17 @@ def _process_decision_events(
...
                     self._apply_input_event(event)
 
                 # Phase 3: Execute workflow logic
                 self._workflow_instance.run_until_yield()
 
-                # If the workflow function returned (or threw an exception), we're done
-                # If it completed early (or late), the nondeterminism tracking will catch that
+                # if get_result throw signal failure, only fail the decision task, not the workflow
+                if self._workflow_instance.get_signal_failure() is not None:
+                    raise self._workflow_instance.get_signal_failure()
```
shijiesheng:
```suggestion
                for signal_name, signal_task in self._workflow_instance.get_signals():
                    if signal_task.done() and not signal_task.cancelled() and (e:=signal_task.exception()) is not None:
                        raise SignalFailure(e, signal_name)
```

cadence/error.py
```diff
@@ -32,14 +32,21 @@ def __init__(self, message: str) -> None:
```
shijiesheng, line 32:
```suggestion

class SignalFailure(Exception):
    def __init__(self, message: str, signal_name: str) -> None:
        super().__init__(message)
        self.signal_name = signal_name
```

cadence/_internal/workflow/workflow_instance.py
timl3136: keeping this as set allow for easy removal of signal task if they complete without issue

cadence/_internal/workflow/workflow_instance.py
timl3136: i dont think this is necessary since we are removing the FIFO ordering into eventloop

cadence/_internal/workflow/context.py
```diff
@@ -30,11 +31,13 @@ def __init__(
         self,
         info: WorkflowInfo,
         decision_manager: DecisionManager,
+        event_loop: DeterministicEventLoop,
```
shijiesheng: not needed. You can use `get_running_loop` because it's already in the loop.

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +61,48 @@ def get_result(self) -> Optional[Payload]:
         if self._task is None or not self._task.done():
             return None
         return self._task.result()
+
+    def handle_signal(self, signal_name: str, payload: Payload) -> None:
+        self._loop.call_soon(self._invoke_signal, signal_name, payload)
```
shijiesheng: simply _invoke_signal(signal_name, payload). There is no need to call_soon here. You already have create_task which will call_soon

cadence/_internal/workflow/workflow_instance.py
```diff
@@ -51,3 +61,48 @@ def get_result(self) -> Optional[Payload]:
...
+            )
+            return
+
+        task = self._loop.create_task(self._run_signal(signal_def, args))
+        self._signal_tasks.add(task)
+        task.add_done_callback(
+            lambda completed_task: self._on_signal_task_done(
+                completed_task, signal_name
+            )
+        )
```
shijiesheng, line 90: Then the last failed signal will be raised rather than the first one?

cadence/client.py
```diff
@@ -452,18 +452,21 @@ def _create_channel(options: ClientOptions) -> Channel:
     interceptors.append(RetryInterceptor())
     interceptors.append(CadenceErrorInterceptor())
 
+    channel_arguments = options.get("channel_arguments") or {}
+    grpc_channel_options: Sequence[tuple[str, Any]] = tuple(channel_arguments.items())
+
```
shijiesheng, line 457: why is the change?

tests/cadence/data_converter_test.py
```diff
@@ -22,31 +22,29 @@ class _TestDataClass:
         pytest.param(
             '"Hello" "world"', [str, str], ["Hello", "world"], id="space delimited"
         ),
-        pytest.param("1", [int, int], [1, 0], id="ints"),
-        pytest.param("1.5", [float, float], [1.5, 0.0], id="floats"),
-        pytest.param("true", [bool, bool], [True, False], id="bools"),
+        pytest.param("1", [int, int], [1], id="ints"),
+        pytest.param("1.5", [float, float], [1.5], id="floats"),
+        pytest.param("true", [bool, bool], [True], id="bools"),
```
shijiesheng: these cases should be supported?

pyproject.toml
```diff
@@ -48,7 +48,8 @@ dev = [
     "pytest-docker>=3.2.3",
     "opentelemetry-instrumentation-grpc==0.60b1",
     "opentelemetry-sdk>=1.39.1",
-    "setuptools-scm[simple]>=9.2"
+    "setuptools-scm[simple]>=9.2",
+    "types-grpcio>=1.0.0.20260408",
```
shijiesheng, line 52: Is this intended?

cadence/client.py
timl3136: it's a type safety change

shijiesheng (approved):

pyproject.toml
shijiesheng: needed by mypy to get the type hints of generated grpc entities

cadence/_internal/workflow/context.py
shijiesheng: mypy

PR #85, opened 2026-04-13
fix(client): compatible with opentelemetry-instrumentation-grpc

shijiesheng:
**What changed?**

* pass in interceptors into insecure and secure channel creation function as keyword argument instead of positional. 
* bug 1 [by AI]: cadence/client.py — Changed from grpc.aio import secure_channel, insecure_channel to import grpc.aio and use grpc.aio.insecure_channel() / grpc.aio.secure_channel() via dot notation. The direct import captured a reference to the original function at import time, so GrpcAioInstrumentorClient's monkey-patching of grpc.aio.insecure_channel had no effect.

* bug 2 [by AI]: cadence/_internal/rpc/retry.py — Changed return await rpc_call to await rpc_call; return rpc_call on success. Returning the raw response caused gRPC to wrap it in UnaryUnaryCallResponse, whose add_done_callback() raises NotImplementedError. Returning the call object preserves the full UnaryUnaryCall interface for outer interceptors like OTel.

**Why?**

incompatible with opentelemetry-instrumentation-grpc==0.60b1 

Issue 1: credits to a Uber employee

When OpenTelemetry gRPC client instrumentation is active, it wraps insecure_channel and injects its own interceptors as a keyword argument. Python receives the interceptors parameter twice — once as positional argument #4 (from the Cadence SDK) and once as a keyword argument (from OpenTelemetry) — and raises TypeError.

Stack Trace

File ".../cadence/client.py", line 123, in __init__
    self._channel = _create_channel(self._options)
File ".../cadence/client.py", line 431, in _create_channel
    return insecure_channel(
File ".../opentelemetry/instrumentation/grpc/__init__.py", line 557, in insecure
    return self._original_insecure(*args, **kwargs)
TypeError: insecure_channel() got multiple values for argument 'interceptors'

Issue 2:
Not Implemented error caused by retryinterceptor (credit to AI)

Stack Trace
File ".../cadence/worker/_poller.py", line 35, in _poll_loop
    await self._poll_and_dispatch()
File ".../cadence/worker/_poller.py", line 44, in _poll_and_dispatch
    task = await self._poll()
File ".../cadence/worker/_activity.py", line 42, in _poll
    await self._client.worker_stub.PollForActivityTask(
File ".../grpc/aio/_interceptor.py", line 471, in __await__
    call = yield from self._interceptors_task.__await__()
File ".../grpc/aio/_interceptor.py", line 701, in _invoke
    call_or_response = await interceptors[0].intercept_unary_unary(
File ".../opentelemetry/instrumentation/grpc/_aio_client.py", line 171, in intercept_unary_unary
    raise NotImplementedError()

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


pyproject.toml
```diff
@@ -46,6 +47,7 @@ dev = [
     "mypy>=1.0.0",
     "pre-commit>=3.0.0",
     "pytest-docker>=3.2.3",
+    "opentelemetry-instrumentation-grpc==0.60b1",
```
timl3136, line 49: do we need to specify this specific version here?

pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "protobuf>=6.31.0,<7.0.0",
     "typing-extensions>=4.0.0",
     "prometheus-client>=0.21.0",
+    "opentelemetry-sdk[dev]>=1.39.1",
```
timl3136: this line should be in the optional-dependencies below

shijiesheng: good catch. uv add --extra dev for some reason moved it here.

pyproject.toml
shijiesheng: yes, this is just to be compatible with what we use internally. We can bump it as it upgrades internally.

timl3136: My question here is that perhaps we can relax the rule to be a range. We don't know if any potential opensource user will have a different version than internally.

shijiesheng: This is beta version anyway and every version upgrade could be breaking. It's reasonable at this stage to ping to an internal version and **only** support that one until the package stablizes.

timl3136 (approved):

PR #87, opened 2026-04-16
chore: separate openai contrib from main type-check

timl3136:
**What changed?**
split core package's ci with plugin ci since min python version is different

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,6 +96,9 @@ exclude = [
     "cadence/api",
     "cadence/api/.*",
     "cadence/sample",
+    # Optional plugins (e.g. OpenAI) use extra deps; omit from default mypy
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -96,6 +96,9 @@ exclude = [
     "cadence/api",
     "cadence/api/.*",
     "cadence/sample",
+    # Optional plugins (e.g. OpenAI) use extra deps; omit from default mypy
+    "cadence/contrib",
+    "cadence/contrib/.*",
```
shijiesheng, line 101: It doesn't look right if we still want `type-check-contrib` step.

PR #95, opened 2026-04-30
feat: active-active support

timl3136:
**What changed?**
Add support for active-active in python client

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


cadence/client.py
```diff
@@ -46,6 +46,61 @@ class StartWorkflowOptions(TypedDict, total=False):
     first_run_at: datetime
     workflow_id_reuse_policy: workflow_pb2.WorkflowIdReusePolicy
     retry_policy: RetryPolicy
+    active_cluster_selection_policy: common_pb2.ActiveClusterSelectionPolicy
+    active_cluster_sticky_region: str
+    active_cluster_external_entity_type: str
+    active_cluster_external_entity_key: str
```
shijiesheng: These are deprecated and are not needed https://github.com/cadence-workflow/cadence-idl/blob/8bcf2d71101b757ee820e2d05249deec9f4dddca/proto/uber/cadence/api/v1/common.proto#L158

shijiesheng: https://github.com/cadence-workflow/cadence-idl/blob/8bcf2d71101b757ee820e2d05249deec9f4dddca/proto/uber/cadence/api/v1/common.proto#L201

c-warren: Yep - we need to pass the cluster_attributes, but the rest of the ActiveClusterSelectionPolicy can be omitted. 
This issue is set up to deprecate the old fields, but you can use it to determine what should be added: https://github.com/cadence-workflow/cadence-idl/issues/262

c-warren: These can be removed, they are deprecated fields

cadence/client.py
```diff
@@ -46,6 +46,61 @@ class StartWorkflowOptions(TypedDict, total=False):
...
+
+
+def _build_active_cluster_selection_policy(
```
c-warren: I think we need to map the cluster attributes here

tests/cadence/test_client_workflow.py
```diff
@@ -458,6 +462,27 @@ def test_invalid_workflow_id_reuse_policy_raises_error(self):
...
 
+    @pytest.mark.asyncio
+    async def test_build_request_with_explicit_active_cluster_selection_policy(
+        self, mock_client
+    ):
+        """Explicit active_cluster_selection_policy should pass through unchanged."""
+        client = Client(domain="test-domain", target="localhost:7933")
+
+        policy = ActiveClusterSelectionPolicy(
+            strategy=ACTIVE_CLUSTER_SELECTION_STRATEGY_REGION_STICKY,
```
c-warren: Let's create a policy with ClusterAttributes to test the happy path - this will be deprecated soon.

c-warren (approved):

PR #99, opened 2026-05-05
feat: define child workflow state machine

timl3136:
**What changed?**
define child workflow state machine

**Why?**
essential step for implement child wf

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
+++ b/cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
@@ -164,11 +164,6 @@ class ChildWorkflowExecutionStateMachine(BaseDecisionStateMachine):
         self._transition(DecisionState.COMPLETED)
         self.result.set_exception(ChildWorkflowExecutionTerminated())
 
-    # RequestCancelExternalWorkflowExecution events reference the child workflow by
-    # workflow_execution.workflow_id (a nested field), not by a bare string id.
-    # The dispatcher resolves dotted paths, so "workflow_execution.workflow_id" extracts
-    # the correct key for the alias lookup.  event_id_is_alias=True registers this event's
-    # ID so that the subsequent handle_cancel_failed can look it up via initiated_event_id.
     @child_workflow_events.event(
         "workflow_execution.workflow_id", event_id_is_alias=True
     )
--- a/cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
+++ b/cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
@@ -21,7 +21,7 @@ from cadence.error import (
     StartChildWorkflowExecutionFailed,
 )
 
-child_workflow_events = EventDispatcher("workflow_id")
+child_workflow_events = EventDispatcher("initiated_event_id")
 
 
 class ChildWorkflowExecutionStateMachine(BaseDecisionStateMachine):
@@ -30,7 +30,7 @@ class ChildWorkflowExecutionStateMachine(BaseDecisionStateMachine):
     request: decision.StartChildWorkflowExecutionDecisionAttributes
     execution: DecisionFuture[WorkflowExecution]
     result: DecisionFuture[Payload]
-    _initiated_event_id: int | None = None
+    _run_id: str | None
 
     def __init__(
         self,
@@ -42,12 +42,10 @@ class ChildWorkflowExecutionStateMachine(BaseDecisionStateMachine):
         self.request = request
         self.execution = execution
         self.result = result
+        self._run_id = None
 
     def get_id(self) -> DecisionId:
         return DecisionId(DecisionType.CHILD_WORKFLOW, self.request.workflow_id)
-
-    def initiated_event_id(self) -> int | None:
-        return self._initiated_event_id
 
     def get_decision(self) -> decision.Decision | None:
         if self.state is DecisionState.REQUESTED:
@@ -56,13 +54,16 @@ class ChildWorkflowExecutionStateMachine(BaseDecisionStateMachine):
             )
         if self.state is DecisionState.CANCELED_AFTER_REQUESTED:
             return record_immediate_cancel(self.request)
-        if self.state is DecisionState.CANCELED_AFTER_RECORDED:
+        if self.state in (
+            DecisionState.CANCELED_AFTER_RECORDED,
+            DecisionState.CANCELED_AFTER_STARTED,
+        ):
             return decision.Decision(
                 request_cancel_external_workflow_execution_decision_attributes=decision.RequestCancelExternalWorkflowExecutionDecisionAttributes(
                     domain=self.request.domain,
                     workflow_execution=WorkflowExecution(
                         workflow_id=self.request.workflow_id,
-                        run_id="",
+                        run_id=self._run_id or "",
                     ),
                     child_workflow_only=True,
                 )
@@ -82,21 +83,18 @@ class ChildWorkflowExecutionStateMachine(BaseDecisionStateMachine):
             return True
 
         if self.state is DecisionState.STARTED:
-            self._transition(DecisionState.CANCELED_AFTER_RECORDED)
+            self._transition(DecisionState.CANCELED_AFTER_STARTED)
             return True
 
         return False
 
-    def set_initiated_event_id(self, initiated_event_id: int) -> None:
-        self._initiated_event_id = initiated_event_id
-
-    @child_workflow_events.event(event_id_is_alias=True)
+    @child_workflow_events.event("workflow_id", event_id_is_alias=True)
     def handle_initiated(
         self, _: history.StartChildWorkflowExecutionInitiatedEventAttributes
     ) -> None:
         self._transition(DecisionState.RECORDED)
 
-    @child_workflow_events.event("initiated_event_id")
+    @child_workflow_events.event()
     def handle_initiation_failed(
         self, event: history.StartChildWorkflowExecutionFailedEventAttributes
     ) -> None:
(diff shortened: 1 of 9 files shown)


cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
```diff
@@ -0,0 +1,178 @@
...
+child_workflow_events = EventDispatcher("workflow_id")
+
+
+class ChildWorkflowExecutionStateMachine(BaseDecisionStateMachine):
+    """State machine for StartChildWorkflowExecution and child close events."""
+
+    request: decision.StartChildWorkflowExecutionDecisionAttributes
+    execution: DecisionFuture[WorkflowExecution]
+    result: DecisionFuture[Payload]
+    _initiated_event_id: int | None = None
```
natemort: Nit: I think this is unused

cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
```diff
@@ -0,0 +1,178 @@
...
+from cadence.api.v1.common_pb2 import Payload, WorkflowExecution
+from cadence.error import (
+    ChildWorkflowExecutionCanceled,
+    ChildWorkflowExecutionFailed,
+    ChildWorkflowExecutionTerminated,
+    ChildWorkflowExecutionTimedOut,
+    StartChildWorkflowExecutionFailed,
+)
+
+child_workflow_events = EventDispatcher("workflow_id")
```
natemort: If you pass `"initiated_event_id"` as the parameter here you won't have to repeat it on all the decorators.

cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
```diff
@@ -0,0 +1,178 @@
...
+            )
+        if self.state is DecisionState.CANCELED_AFTER_REQUESTED:
+            return record_immediate_cancel(self.request)
+        if self.state is DecisionState.CANCELED_AFTER_RECORDED:
+            return decision.Decision(
+                request_cancel_external_workflow_execution_decision_attributes=decision.RequestCancelExternalWorkflowExecutionDecisionAttributes(
+                    domain=self.request.domain,
+                    workflow_execution=WorkflowExecution(
+                        workflow_id=self.request.workflow_id,
+                        run_id="",
```
natemort: This is an interesting case. We don't have a RunID (as it hasn't started yet) but we'd like to cancel it. Does this match how the other clients handle it?

Since transfer tasks can be executed in parallel I wonder if this would create a race condition on the server. We might try canceling the Workflow before it has ever started. We might need to wait for it to start before we can cancel it.

cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
```diff
@@ -0,0 +1,178 @@
...
+        return self._initiated_event_id
+
+    def get_decision(self) -> decision.Decision | None:
+        if self.state is DecisionState.REQUESTED:
+            return decision.Decision(
+                start_child_workflow_execution_decision_attributes=self.request
+            )
+        if self.state is DecisionState.CANCELED_AFTER_REQUESTED:
+            return record_immediate_cancel(self.request)
+        if self.state is DecisionState.CANCELED_AFTER_RECORDED:
```
natemort: I think we also need to model the CANCELED_AFTER_STARTED scenario to cancel an in-flight Child workflow. At that point we'd have the RunID.

cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
```diff
@@ -0,0 +1,178 @@
...
+            self.result.force_cancel()
+            return True
+
+        if self.state is DecisionState.RECORDED:
+            self._transition(DecisionState.CANCELED_AFTER_RECORDED)
+            self.execution.force_cancel()
+            return True
+
+        if self.state is DecisionState.STARTED:
+            self._transition(DecisionState.CANCELED_AFTER_RECORDED)
```
natemort: Nit: I mentioned this above, but this may be a separate state because the Workflow has already started.

cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+
+    @child_workflow_events.event("initiated_event_id")
+    def handle_terminated(
+        self, event: history.ChildWorkflowExecutionTerminatedEventAttributes
+    ) -> None:
+        self._transition(DecisionState.COMPLETED)
+        self.result.set_exception(ChildWorkflowExecutionTerminated())
+
+    @child_workflow_events.event("workflow_execution")
```
natemort: The state machine is never keyed by the WorkflowExecution, only the WorkflowID. To dispatch this event it would grab the value of the `worklfow_execution` field and look it up in the `aliases` dict in `DecisionManager`.

We likely need to extend the EventDispatcher and/or DecisionManager to handle this case.

Oh this is what gitar is commenting on below.

cadence/_internal/workflow/statemachine/nondeterminism.py
```diff
@@ -267,6 +267,33 @@ def _(attrs: history.ActivityTaskCancelRequestedEventAttributes) -> Expectation:
...
+    attrs: history.StartChildWorkflowExecutionInitiatedEventAttributes,
+) -> Expectation:
+    return Expectation(
+        DecisionId(DecisionType.CHILD_WORKFLOW, attrs.workflow_id),
+        {"workflow_type": attrs.workflow_type.name},
+    )
+
+
+@to_expectation.register
+def _(attrs: history.StartChildWorkflowExecutionFailedEventAttributes) -> Expectation:
```
natemort, line 289: The expectation value shouldn't be CANCEL here and we need to add Expectation building for the other cancellation events.

cadence/error.py
```diff
@@ -144,6 +126,35 @@ class QueryFailedError(CadenceRpcError):
 
 
 class ServiceBusyError(CadenceRpcError):
-    def __init__(self, message: str | None, code: grpc.StatusCode, reason: str) -> None:
+    def __init__(self, message: str, code: grpc.StatusCode, reason: str) -> None:
         super().__init__(message, code, reason)
         self.reason = reason
+
+
+class StartChildWorkflowExecutionFailed(Exception):
```
natemort: We probably want these to form a hierarchy to make it easier to handle, and potentially consolidate them.

cadence/_internal/workflow/statemachine/child_workflow_execution_state_machine.py
```diff
@@ -0,0 +1,179 @@
...
+        if self.state is DecisionState.REQUESTED:
+            return decision.Decision(
+                start_child_workflow_execution_decision_attributes=self.request
+            )
+        if self.state is DecisionState.CANCELED_AFTER_REQUESTED:
+            return record_immediate_cancel(self.request)
+        if self.state in (
+            DecisionState.CANCELED_AFTER_RECORDED,
+            DecisionState.CANCELED_AFTER_STARTED,
+        ):
```
shijiesheng: nit: use switch in python `match`

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -152,8 +173,9 @@ def handle_history_event(self, event: history.HistoryEvent) -> None:
             decision_type = event_action.decision_type
             action = event_action.action
             # Find what state machine the event references.
-            # This may be a reference via the user id or a reference to a previous event
-            id_for_event = getattr(event_attributes, action.id_attr)
+            # This may be a reference via the user id or a reference to a previous event.
+            # Supports dotted paths (e.g. "workflow_execution.workflow_id") for nested fields.
+            id_for_event = resolve_id_attr(event_attributes, action.id_attr)
```
shijiesheng: just a comment: we might want to revisit to simplify the logic. The complexity is to point an event to a statemachine id. 
In the past, it's simply resolved by alias logic. Now with childworkflow, the mapping logic becomes more complicated.

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -110,6 +116,21 @@ def start_timer(
...
+        if self._replaying:
+            self._determinism_tracker.validate_action(attrs)
+        decision_id = DecisionId(DecisionType.CHILD_WORKFLOW, attrs.workflow_id)
+        execution: DecisionFuture[WorkflowExecution] = self._create_future(decision_id)
+        result: DecisionFuture[Payload] = DecisionFuture(
+            self._event_loop, lambda: self._request_cancel(decision_id)
+        )
+        machine = ChildWorkflowExecutionStateMachine(attrs, execution, result)
+        self._add_state_machine(machine)
+        return execution, result
```
shijiesheng, line 134: why do we need the result here?

cadence/error.py
```diff
@@ -147,3 +147,38 @@ class ServiceBusyError(CadenceRpcError):
...
+        self.details = details
+
+
+class ChildWorkflowExecutionTimedOut(ChildWorkflowError):
+    def __init__(self, message: str, timeout_type: int) -> None:
+        super().__init__(message)
+        self.timeout_type = timeout_type
+
+
+class ChildWorkflowExecutionTerminated(ChildWorkflowError):
```
shijiesheng: All errors in this file is exposed to `Client`. We don't have APIs on child workflows and thus we don't need it here. The suggestion would be to move to where the error is needed.

idls
natemort, line 1: I don't think you intended to update this

natemort (approved):
LGTM, +1 to Shijie's comments.

PR #103, opened 2026-05-12
feat(query): add query decorator and client method of query_workflow

shijiesheng:
**What changed?**

* decorator and definition
* client method of query_workflow

**Why?**

Query Feature

**How did you test it?**

Interface change, unit test and integration test will come in the next PRs

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/workflow.py
+++ b/cadence/workflow.py
@@ -22,6 +22,7 @@ import inspect
 from cadence._internal.fn_signature import FnSignature
 from cadence.data_converter import DataConverter
 from cadence.error import ContinueAsNewError
+from cadence.query import QueryDefinition, QueryDefinitionOptions
 from cadence.signal import SignalDefinition, SignalDefinitionOptions
 
 ResultType = TypeVar("ResultType")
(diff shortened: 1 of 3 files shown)


cadence/workflow.py
```diff
@@ -338,6 +363,47 @@ def decorator(f: T) -> T:
...
+
+        @workflow.query(name="get_count")
+        def get_count(self, prefix: str) -> int:
+            return self.counts.get(prefix, 0)
+
+    Constraints:
+        * Query handlers must have a non-None return type.
+        * Query handlers must be synchronous (not async).
+        * Query handlers must not mutate workflow state.
+        * A method can only be one of @workflow.run, @workflow.signal, or @workflow.query.
```
timl3136, line 388: i think this comment is wrong, we should allow for non-decorated function in a workflow class

cadence/workflow.py
```diff
@@ -338,6 +363,47 @@ def decorator(f: T) -> T:
...
+
+    Returns:
+        The decorated method with workflow query metadata
+
+    Raises:
+        ValueError: If name is not provided
+    """
+
+    def decorator(f: T) -> T:
+        f._workflow_query = name or f.__qualname__  # type: ignore
```
timl3136: nit: type: ignore

timl3136: We are not returning valueError if name is not provided here

`if name is None:
        raise ValueError("name is required")`

timl3136 (commented):
codecov is failing due to missing unit test

cadence/workflow.py
shijiesheng: Oh, it actually meant we do not allow a method to be both run and signal. It says "can" not "is".

cadence/workflow.py
shijiesheng: I'm planning to change the behavior. We should allow registering without name and fallback to method name directly.

timl3136: perhaps we can reject for now, then change the behavior for all decorators in a separate PR for consistency.

timl3136 (approved):

cadence/workflow.py
shijiesheng: fixed

shijiesheng: sure

PR #108, opened 2026-05-15
test(schedule): add integration tests for schedule client API

abhishekj720:
**What changed?**
One new fileL tests/integration_tests/test_schedule.py. 

**Why?**
Validates the 8 schedule RPCs against a real Cadence server. 
Unit tests use a fake servicer and can't catch proto field mapping bugs or server-version incompatibilities. Those are simple test to start a schedule and describe it.

**How did you test it?**
docker compose -f tests/integration_tests/docker-compose.yml up -d
uv run pytest tests/integration_tests/test_schedule.py --integration-tests -v

**Potential risks**
- test_list_schedules_contains_created could be flaky if the domain has thousands of schedules and pagination cuts off before reaching the new one (default page_size=100). In a clean CI environment this won't happen.

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/tests/integration_tests/test_schedule.py
+++ b/tests/integration_tests/test_schedule.py
@@ -1,0 +1,166 @@
+"""Integration tests for schedule client methods.
+
+Requires a running Cadence server. Run with:
+    uv run pytest tests/integration_tests/test_schedule.py --integration-tests -v
+"""
+
+import asyncio
+import time
+import uuid
+
+import pytest
+from google.protobuf import text_format
+
+from cadence.api.v1 import common_pb2, schedule_pb2, tasklist_pb2
+from cadence.api.v1.service_schedule_pb2 import DescribeScheduleResponse
+from cadence.client import Client
+from cadence.error import QueryFailedError
+from tests.integration_tests.helper import CadenceHelper
+
+
+def _dummy_schedule_action() -> schedule_pb2.ScheduleAction:
+    """Cadence rejects CreateSchedule without Action.start_workflow set."""
+    return schedule_pb2.ScheduleAction(
+        start_workflow=schedule_pb2.ScheduleAction.StartWorkflowAction(
+            workflow_type=common_pb2.WorkflowType(
+                name="ScheduleIntegrationDummyWorkflow"
+            ),
+            task_list=tasklist_pb2.TaskList(name="schedule-integration-task-list"),
+        )
+    )
+
+
+async def _describe_schedule_when_ready(
+    client: Client,
+    schedule_id: str,
+    *,
+    timeout_s: float = 90.0,
+) -> DescribeScheduleResponse:
+    """DescribeSchedule is backed by a query; the schedule workflow may not be queryable immediately."""
+    deadline = time.monotonic() + timeout_s
+    backoff = 0.25
+    last_err: QueryFailedError | None = None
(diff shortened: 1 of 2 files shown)


tests/integration_tests/test_schedule.py
```diff
@@ -0,0 +1,165 @@
...
+    async with helper.client() as client:
+        try:
+            await client.create_schedule(
+                schedule_id,
+                spec=schedule_pb2.ScheduleSpec(cron_expression="0 11 * * *"),
+                action=_dummy_schedule_action(),
+            )
+
+            ids = [e.schedule_id async for e in client.list_schedules()]
+            assert schedule_id in ids
```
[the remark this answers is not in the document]
abhishekj720: We list immediately after create; list_schedules paginates fully. Describe-before-list would only help if we observe list lagging behind create in CI.

tests/integration_tests/test_schedule.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+    if resp.action != expected_action:
+        raise AssertionError(
+            "DescribeSchedule action mismatch:\n"
+            f"--- expected ---\n{text_format.MessageToString(expected_action)}\n"
+            f"--- got ---\n{text_format.MessageToString(resp.action)}"
+        )
+
+
+@pytest.mark.usefixtures("helper")
```
shijiesheng, line 78: nit: not needed actually. could you help clean up the other files as well.

tests/integration_tests/test_schedule.py
```diff
@@ -0,0 +1,165 @@
...
+
+from cadence.api.v1 import common_pb2, schedule_pb2, tasklist_pb2
+from cadence.api.v1.service_schedule_pb2 import DescribeScheduleResponse
+from cadence.client import Client
+from cadence.error import QueryFailedError
+from tests.integration_tests.helper import CadenceHelper
+
+
+def _dummy_schedule_action() -> schedule_pb2.ScheduleAction:
+    """Cadence rejects CreateSchedule without Action.start_workflow set."""
```
shijiesheng: it's not a dummy at all. It will run a workflow

tests/integration_tests/test_schedule.py
```diff
@@ -0,0 +1,165 @@
...
+    timeout_s: float = 90.0,
+) -> DescribeScheduleResponse:
+    """DescribeSchedule is backed by a query; the schedule workflow may not be queryable immediately."""
+    deadline = time.monotonic() + timeout_s
+    backoff = 0.25
+    last_err: QueryFailedError | None = None
+    while time.monotonic() < deadline:
+        try:
+            return await client.describe_schedule(schedule_id)
+        except QueryFailedError as exc:
```
shijiesheng, line 46: Do we actually hit this QueryFailedError at all? Also it's not very intuitive to users as why it's QueryFailedError when they are trying to get the schedule. It's an implementation detail that could be handled.

shijiesheng (approved):

tests/integration_tests/test_schedule.py
abhishekj720: Will move the retry into client.describe_schedule internally, then remove _describe_schedule_when_ready from the test entirely in a follow up PR

PR #114, opened 2026-05-21
feat: add execute_child_workflow

timl3136:
**What changed?**
Add StartChildWorkflowOptions and execute_child_workflow (This will block the thread until child wf completes)

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/workflow/context.py
+++ b/cadence/_internal/workflow/context.py
@@ -7,15 +7,18 @@ from typing import Iterator, Optional, Any, Unpack, Type, cast, Callable
 from cadence._internal.workflow.deterministic_event_loop import DeterministicEventLoop
 from cadence._internal.workflow.retry_policy import retry_policy_to_proto
 from cadence._internal.workflow.statemachine.decision_manager import DecisionManager
-from cadence.api.v1.common_pb2 import ActivityType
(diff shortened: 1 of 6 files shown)


cadence/_internal/workflow/context.py
```diff
@@ -105,6 +108,75 @@ async def execute_activity(
...
+            retry_policy=retry_policy_to_proto(kwargs.get("retry_policy")),
+        )
+
+        cron_schedule = kwargs.get("cron_schedule")
+        if cron_schedule:
+            schedule_attributes.cron_schedule = cron_schedule
+
+        _execution_future, result_future = (
+            self._decision_manager.schedule_child_workflow(schedule_attributes)
+        )
```
shijiesheng, line 173: AI also suggests that `schedule_child_workflow` only requires one return. You could just return the execution future. The result can be accessed directly from there

timl3136: We are returning execution and result future in sync with Java client. I don't think it would make a big difference here.

shijiesheng: discussed offline, this might be used for signals. So let's keep it for now.

shijiesheng (approved):

PR #115, opened 2026-05-22
test: harden schedule backfill polling in CI

abhishekj720:
**What changed?**

Updated the schedule backfill integration test to better tolerate transient server behavior in CI. The test now keeps polling when `DescribeSchedule` briefly returns `DEADLINE_EXCEEDED`, extends the polling window from 30s to 60s, and documents why transient polling errors are ignored during backfill processing.

**Why?**

The failing GitHub Actions `Unit Tests` job was caused by `tests/integration_tests/test_schedule.py::test_backfill` failing intermittently while the scheduler was still processing the backfill request. The Actions logs showed `describe_schedule()` timing out with `Deadline Exceeded`, so this change makes the test resilient to that transient condition without changing production behavior.

**How did you test it?**

- Reviewed the failing Actions logs for job `80146212522`
- Ran `uv tool run ruff check tests/integration_tests/test_schedule.py`
- Ran `uv run pytest -v tests/cadence/test_schedule.py` (`21 passed`)
- Attempted to run `uv run pytest -v tests/integration_tests/test_schedule.py::test_backfill --integration-tests`, but the local Docker/Cassandra test environment hung on startup due to an unrelated sandbox DNS/container issue

**Potential risks**

Low risk. This is a test-only change and does not modify runtime client behavior. The main impact is that a genuinely failing backfill test may now take longer to fail because of the longer polling window.

**Release notes**

None. Test-only change.

**Documentation Changes**

None.


tests/integration_tests/test_schedule.py
```diff
@@ -140,3 +153,44 @@ async def test_list_schedules_contains_created(helper: CadenceHelper):
...
+@pytest.mark.usefixtures("helper")
+async def test_backfill(helper: CadenceHelper):
+    """Backfill starts workflow executions for past schedule slots."""
+    schedule_id = f"test-schedule-backfill-{uuid.uuid4()}"
+
+    async with helper.client() as client:
+        try:
+            await client.create_schedule(
+                schedule_id,
+                spec=schedule_pb2.ScheduleSpec(cron_expression="* * * * *"),
```
YaweiZhang-930, line 174: It's worth noting that **total_runs** doesn't distinguish **backfill** fires from **regular cron fires** (server increments the same counter for both). So with * * * * *(fire every minute) and a 30s deadline, roughly half the time the next minute boundary falls inside our poll window and a normal cron tick alone satisfies >= 1. The test would still pass if backfill silently produced nothing.

PR #116, opened 2026-05-26
fix: skip_if_schedule_api_unimplemented due to outdated docker image

timl3136:
**What changed?**
outdated docker image on schedule causing unit test to fail, we want to skip the unimplemented error in the meantime

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tests/integration_tests/test_schedule.py
+++ b/tests/integration_tests/test_schedule.py
@@ -8,13 +8,14 @@ import asyncio
 import time
 import uuid
 
+import grpc
 import pytest
 from google.protobuf import text_format
 
 from cadence.api.v1 import common_pb2, schedule_pb2, tasklist_pb2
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_schedule.py
```diff
@@ -96,7 +122,7 @@ async def test_create_describe_delete(helper: CadenceHelper):
             await client.delete_schedule(schedule_id)
 
 
-@pytest.mark.usefixtures("helper")
+@pytest.mark.usefixtures("helper", "schedule_api_available")
```
shijiesheng, line 125: I don't think you'll need this line at all. How is "schedule_api_availabe" used? For helper, it's not needed because it's already in the argument.

timl3136:
Root cause was my local machine's stale docker image.

PR #122, opened 2026-05-28
feat(workflow): add SignalExternalWorkflowStateMachine

timl3136:
**What changed?**
add SignalExternalWorkflowStateMachine, only state machine, no wiring yet

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/workflow/statemachine/signal_external_workflow_state_machine.py
+++ b/cadence/_internal/workflow/statemachine/signal_external_workflow_state_machine.py
@@ -52,7 +52,6 @@ class SignalExternalWorkflowStateMachine(BaseDecisionStateMachine):
     def request_cancel(self) -> bool:
         return False
 
-    @signal_external_events.event(id_attr="control", event_id_is_alias=True)
     def handle_initiated(
         self,
         _: history.SignalExternalWorkflowExecutionInitiatedEventAttributes,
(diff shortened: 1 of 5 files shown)


cadence/_internal/workflow/statemachine/signal_external_workflow_state_machine.py
```diff
@@ -0,0 +1,81 @@
...
+        self.completed.set_result(None)
+
+    @signal_external_events.event()
+    def handle_failed(
+        self,
+        event: history.SignalExternalWorkflowExecutionFailedEventAttributes,
+    ) -> None:
+        self._transition(DecisionState.COMPLETED)
+        self.completed.set_exception(
+            SignalExternalWorkflowFailed(
```
shijiesheng, line 77: We also have SignalFailure that's surfaced to the user. Do we want to let users to catch this error inside workflow code? If so, we will need to somehow wrap it as SignalFailure as well.

timl3136: those two are different, signalFailure is when workflow cannot handle the incoming signal, the exception here is a outbound failure where server rejects the request

shijiesheng (approved):

PR #124, opened 2026-05-29
feat: incorporate SignalExternalWorkflowStateMachine into DecisionManager

timl3136:
**What changed?**
- Added signal_external_workflow to DecisionManager
- Registered signal external workflow

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/workflow/statemachine/nondeterminism.py
+++ b/cadence/_internal/workflow/statemachine/nondeterminism.py
@@ -344,6 +344,17 @@ def _(
     )
 
 
+@to_expectation.register
+def _(
+    attrs: history.SignalExternalWorkflowExecutionFailedEventAttributes,
+) -> Expectation:
+    # signal_name is not present in failed attrs; control still identifies the decision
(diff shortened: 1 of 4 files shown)


shijiesheng (commented):
Could you add an integration test?

cadence/_internal/workflow/statemachine/nondeterminism.py
```diff
@@ -323,6 +323,27 @@ def _(
     )
 
 
+# Signal External Workflow - Enforce signal_name
+@to_expectation.register
```
shijiesheng, line 327: Do we need to add SignalExternalWorkflowExecutionFailedEventAttributes as well?

shijiesheng (approved):

cadence/_internal/workflow/statemachine/nondeterminism.py
timl3136: Yes, i will update that here

PR #126, opened 2026-06-02
feat: implement signal_external_workflow and signal API call for child wf

timl3136:
**What changed?**
- Added workflow APIs to send signals via Cadence’s **SignalExternalWorkflowExecution** decision
- Added **`signal_child_workflow()`** on the internal context

**Why?**
those are important feature

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/workflow/context.py
+++ b/cadence/_internal/workflow/context.py
@@ -10,9 +10,10 @@ from cadence._internal.workflow.deterministic_event_loop import DeterministicEve
 from cadence._internal.workflow.retry_policy import retry_policy_to_proto
 from cadence._internal.workflow.statemachine.decision_manager import DecisionManager
 from cadence.api.v1 import workflow_pb2
-from cadence.api.v1.common_pb2 import ActivityType, WorkflowType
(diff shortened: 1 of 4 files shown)


cadence/_internal/workflow/context.py
```diff
@@ -244,6 +271,30 @@ def _activate(self) -> Iterator["Context"]:
...
+            domain=domain or self._info.workflow_domain,
+            workflow_execution=WorkflowExecution(
+                workflow_id=workflow_id,
+                run_id=run_id,
+            ),
+            signal_name=signal_name,
+            input=self.data_converter().to_data(list(args)),
+            child_workflow_only=child_workflow_only,
+        )
+        await self._decision_manager.signal_external_workflow(attrs)
```
shijiesheng, line 296: I just realized that decision manager methods are always sync method that returns future directly. It's not very pythonic. Instead, it's better to use async method.

shijiesheng: Maybe we can fix as a followup

timl3136: I will create a follow up PR to address this.

shijiesheng (approved):

PR #128, opened 2026-06-04
feat: implement workflow cancellation

timl3136:
**What changed?**
implement workflow cancellation

**Why?**
Cancellation is a important feature that allows for gracefully exit

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


cadence/_internal/workflow/workflow_engine.py
```diff
@@ -308,6 +323,16 @@ def _handle_signaled_input_event(
...
         self._workflow_instance.handle_signal(signal_def, args)
 
+    @_handle_input_event.register
+    def _handle_cancel_requested_input_event(
+        self,
+        attrs: WorkflowExecutionCancelRequestedEventAttributes,
+        event: HistoryEvent,
+    ) -> None:
+        info = self._context.request_cancel(attrs)
+        self._decision_manager.request_cancel_pending_decisions(info.cause)
```
natemort: I don't think this (specifically cancelling all pending decision state machines) is the correct behavior, but I'm not positive on the correct solution.

Working backwards from the behavior we want to support, users should be able to:
- Ignore cancellation entirely for some specific action. Java/Go refer to this as being a disconnected context IIRC. The analogy in Python seems to be `asyncio.shield`.
- Respond to the cancellation and run other actions. For example, you might catch the exception an Activity throws and then run a cleanup operation within a new disconnected context.
- Do nothing and get cancelled.

The other thing is understanding the standard Python semantics around cancellation and trying to appear as "native" as possible. Our Java SDK mostly fails on this since Java doesn't have a good idiom, but our Go SDK uses the explicit context propagation and cancellation that you would use in "normal" Go code.

From my limited Python understanding, when we cancel a task (done on the next line down in `workflow_instance.request_cancel`) any time that coroutine attempts to interact with another Future it'll cancel that future.  So if a user has code like this:
```
a = workflow.execute_activity(...)
// gets cancelled
await a // throws CancelledError
```
then it works fine. But if they have code like this:
```
a = workflow.execute_activity(...)
b = workflow.execute_activity(...)
// gets cancelled
await a // throws CancelledError
await b
```
then we don't cancel the second activity like we would in Java/Go. So we definitely need something more.

It seems like the "Pythonic" thing is using a TaskGroup, but that's explicit structured concurrency and doesn't match our existing patterns. We could maybe introduce our own implicit task grouping and cancellation via ContextVars or something like that

timl3136: Thanks

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -142,7 +144,8 @@ def schedule_child_workflow(
         execution: DecisionFuture[WorkflowExecution] = self._create_future(decision_id)
         execution.add_done_callback(_consume_future_exception)
         result: DecisionFuture[Payload] = DecisionFuture(
-            self._event_loop, lambda: self._request_cancel(decision_id)
+            self._event_loop,
+            self._create_cancel_callback(decision_id),
         )
```
shijiesheng: nit: 
```suggestion
        result: DecisionFuture[Payload] = self._create_future(decision_id)
```

cadence/_internal/workflow/workflow_engine.py
```diff
@@ -224,7 +226,20 @@ def _maybe_complete_workflow(self) -> Optional[Decision]:
                     result=self._data_converter.to_data([result]),
                 )
             )
-        except (CancelledError, InvalidStateError, FatalDecisionError):
+        except CancelledError as e:
+            if not self._context.is_cancel_requested():
+                raise
```
shijiesheng: Same as activity, we should allow users to cancel their own workflow by throwing cancelled exceptions.

cadence/_internal/workflow/statemachine/decision_state_machine.py
```diff
@@ -63,7 +63,7 @@ def get_id(self) -> DecisionId: ...
 
     def get_decision(self) -> decision.Decision | None: ...
 
-    def request_cancel(self) -> bool: ...
+    def request_cancel(self, message: str | None = None) -> bool: ...
```
shijiesheng, line 67: We never ever need this in the interface. In each state machine, force_cancel is just used internally and should be private

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -255,23 +258,43 @@ def _end_execution(self) -> None:
...
+            if d is not None:
+                decisions.append(d)
 
-        for machine in self.state_machines.values():
-            to_send = machine.get_decision()
-            if to_send is not None:
-                decisions.append(to_send)
+        if is_completing:
+            assert completion_decision is not None  # same predicate as is_completing
+            decisions.append(completion_decision)
```
shijiesheng, line 274: nit: These can be simplified if you change the complete_state_machine to workflow state machine that supports cancel. Workflow cancel should still yield decisions like workflow completed/cancelled/failed event.

cadence/_internal/workflow/context.py
```diff
@@ -268,6 +271,22 @@ async def wait_condition(self, predicate: Callable[[], bool]) -> None:
...
+            cause=attrs.cause,
+            identity=attrs.identity,
+            request_id=attrs.request_id,
+        )
+        return self._cancellation_info
+
+    def is_cancel_requested(self) -> bool:
+        return self._cancellation_info is not None
+
+    def cancellation_info(self) -> WorkflowCancellationInfo | None:
```
shijiesheng: nit: I guess We don't need this for now.

cadence/_internal/workflow/statemachine/decision_manager.py
timl3136: Thank you

cadence/_internal/workflow/workflow_engine.py
timl3136: Thanks, added this logic as well as unit test

cadence/_internal/workflow/statemachine/decision_manager.py
timl3136: Thanks, this logic is already simplified when we remove request_cancel_pending_decisions and corresponding logic.

shijiesheng (approved):

PR #130, opened 2026-06-08
feat(schedule): add lifecycle sample and update_schedule concurrency note

abhishekj720:
## Summary

- **`cadence/sample/schedule_example.py`** : new sample demonstrating the full schedule lifecycle: create, describe, pause, unpause, update (read-modify-write), list, backfill, and delete. Includes an embedded worker so workflows triggered by the schedule actually complete.
- **`cadence/client.py`** : adds a concurrency warning to the `update_schedule` docstring: concurrent updates to the same schedule ID are not safe (last-write-wins, no conflict token exposed by the server API).

## Test plan

- [ ] `uv run pytest tests/cadence/ -q`
- [ ] Manual: start docker-compose, run `uv run python cadence/sample/schedule_example.py`, verify each step prints without error

### DIFF
--- a/cadence/client.py
+++ b/cadence/client.py
@@ -680,6 +680,9 @@ class Client:
 
         Fetches the current schedule state, passes it to ``updater`` for
         modification, then sends the full updated state to the server.
+
+        Note: concurrent updates to the same schedule_id are not safe —
+        last write wins. The server API exposes no conflict token.
         """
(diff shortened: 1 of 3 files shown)


shijiesheng:
moved the sample here https://github.com/cadence-workflow/cadence-samples/pull/155

PR #133, opened 2026-06-10
feat: activity cancellation

timl3136:
**What changed?**
Implement activity cancellation

**Why?**
When a cancellation request sent to the server, and the activity have heartbeat, the next heartbeat will inform the worker that the activity is pending for cancellation which allows custom logic to be run before cancelling the execution of activity.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


shijiesheng (commented):
I don't think _ActivityCancellation is needed. For async activities, asyncio already handles cancellation natively.

shijiesheng (commented):
https://github.com/shijiesheng/cadence-python-client/commit/84346180a061a5791f9f2794b9e46c1fc6573b9b

cadence/_internal/activity/_context.py
```diff
@@ -58,6 +73,9 @@ def heartbeat(self, *details: Any) -> None:
     def heartbeat_details(self, *types: Type) -> list[Any]:
         return self._heartbeat_sender.get_details(*types)
 
+    def is_cancelled(self) -> bool:
```
shijiesheng, line 100: We might also need APIs like wait_for_cancelled(timeout: timedelta = None) for both sync and async

cadence/_internal/activity/_activity_executor.py
```diff
@@ -38,11 +40,18 @@ def __init__(
...
+        context: Union[_Context, _SyncContext] | None = None
         try:
             context = self._create_context(task)
             result = await context.execute(task.input)
             await self._report_success(task, result)
+        except ActivityCancelledError as e:
+            if context is not None and context.is_cancelled():
+                await self._report_cancelled(task, e.details)
+                return
+            _logger.exception("Activity failed")
```
shijiesheng: It's a genuine library bug rather than an actual activity failure. Maybe error message should mention that. If possible, we can also add a InternalError to indicate it's a library fault. 
```suggestion
            _logger.exception(
    "Activity failed: ActivityCancelledError raised but the activity was not cancelled; this is a bug in cadence-python-client"
)
```

cadence/_internal/activity/_context.py
```diff
@@ -93,3 +121,13 @@ def heartbeat(self, *details: Any) -> None:
...
         wrapped.add_done_callback(self._heartbeat_tasks.discard)
+
+    def wait_for_cancelled(self, timeout: timedelta | float | None = None) -> bool:
+        if timeout is None:
+            sec: float | None = None
+        elif isinstance(timeout, timedelta):
+            sec = timeout.total_seconds()
+        else:
+            sec = float(timeout)
+        return self._heartbeat_sender.wait_for_cancellation(sec)
```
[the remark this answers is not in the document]
timl3136: it should block forever if not specified, that's by design. It will timeout eventually.

cadence/activity.py
```diff
@@ -79,6 +79,28 @@ def heartbeat_details(*types: Type) -> list[Any]:
     return ActivityContext.get().heartbeat_details(*types)
 
 
+def is_cancelled() -> bool:
+    """Return whether cancellation has been requested for the current activity."""
+    return ActivityContext.get().is_cancelled()
+
+
+def wait_for_cancelled(timeout: timedelta | float | None = None) -> bool:
```
natemort, line 133: What's the analogy for this in the other clients? I'm not sure how users would use this.

cadence/_internal/activity/_context.py
```diff
@@ -21,16 +23,28 @@ def __init__(
...
         self._heartbeat_tasks: set[asyncio.Future[None]] = set()
 
     async def execute(self, payload: Payload) -> Any:
         params = self._to_params(payload)
+        self._activity_task = asyncio.create_task(self._run_activity(params))
         try:
-            with self._activate():
-                return await self._activity_def.impl_fn(*params)
+            result = await self._activity_task
+            if self.is_cancelled():
```
natemort, line 38: Activities may ignore cancellation. I think the following is possible:
- Activity heartbeats
- ActivityContext sees cancellation_requested and is_cancelled() is now True.
- Activity returns a value

In that case we want to report the value and complete the activity. This would be very common for sync activities since we can't cancel them as easily/obviously.

Particularly with the `wait_pending_heartbeats` logic it would be possible for their activity function to return before the activity is cancelled but then we find out it was cancelled after their logic finishes executing.

cadence/_internal/activity/_heartbeat.py
```diff
@@ -23,20 +24,29 @@ def __init__(
         self._task_token = task_token
         self._identity = identity
         self._previous_details = previous_details
+        self._cancel_requested = threading.Event()
```
natemort: Should be asyncio.Event

cadence/_internal/activity/_context.py
```diff
@@ -21,16 +23,28 @@ def __init__(
...
+                raise ActivityCancelledError()
+            return result
+        except asyncio.CancelledError:
```
natemort: This should be the main mechanism for detecting Cancellation.
- If their activity raises the CancelledError and we requested cancellation, then we're cancelled.
- If they otherwise raise it then it's a failure with a reason of CancelledError.

I think it would be reasonable to accept either CancelledError or ActivityCancelledError as the exceptions they can raise when cancelled because `asyncio.CancelledError` would look weird in a sync activity impl.

cadence/activity.py
natemort, line 55: It might be useful to make `ActivityContext.get()` raise `ActivityCancelledError` if the activity is cancelled. Sync activities would be way more likely to report cancellation and stop processing since they'll presumably interact with the Activity API occasionally (especially for Logging/Metrics).

What do you think?

cadence/_internal/activity/_heartbeat.py
shijiesheng: HeartbeatSender should be stateless. Move the logic up to the Context. Same for wait_for_cancellation method

cadence/_internal/activity/_context.py
shijiesheng: +1, heartbeat actually returns workflow's decision to "request" cancel. Activity should be the one to report its final result correctly.

cadence/activity.py
timl3136: one analogy I can think of for Go is like following: https://github.com/cadence-workflow/cadence-go-client/blob/66a05010c99051ebb99ba3689ca7affe60f7c3fd/internal/internal_workflow_testsuite_test.go#L549

that allows user to do work before raise the cancelledError

cadence/_internal/activity/_heartbeat.py
timl3136: Thanks

cadence/_internal/activity/_context.py
timl3136: Thanks, I have made the change here

cadence/activity.py
timl3136: I think that's a good idea, let me add that

shijiesheng (approved):

PR #138, opened 2026-06-15
fix(error): add CadenceRpcError.message attribute

abhishekj720:
## Summary

Adds `message` as a named attribute on `CadenceRpcError` so callers have a
clean way to access the error message without reaching into `Exception.args[]`.

## What this PR does NOT do

The scheduler mid-ContinueAsNew error (`StatusCode.UNKNOWN` with message
`"cadence internal uncategorized error, msg: code:unavailable..."`) is a
**server-side bug**: `metered.go`'s `handleErr` only passes through
`CodeDeadlineExceeded` and `CodeCancelled` from `*yarpcerrors.Status` —
`CodeUnavailable` falls through and gets wrapped as a plain `fmt.Errorf`,
which `FromError` then maps to `CodeUnknown`.

The correct fix is server-side:
1. `service/frontend/wrappers/metered/metered.go`: pass through `CodeUnavailable`
   alongside `CodeDeadlineExceeded` and `CodeCancelled`
2. `common/types/mapper/proto/errors.go` `FromError`: handle `*yarpcerrors.Status`
   by preserving its code rather than always falling through to `CodeUnknown`

Once that is fixed, `UNAVAILABLE` is already in the client's `RETRYABLE_CODES`
and the retry happens automatically — no client change is needed.

## Test plan

- [x] `uv run pytest tests/cadence/_internal/rpc/test_retry.py -v` — 14 tests pass
- [x] `uv run mypy cadence/` — no errors
- [x] `uv run ruff check && ruff format --check` — clean

### DIFF
--- a/cadence/_internal/rpc/retry.py
+++ b/cadence/_internal/rpc/retry.py
@@ -113,14 +113,4 @@ def is_retryable(err: CadenceRpcError, call_details: ClientCallDetails) -> bool:
             and err.active_cluster != err.current_cluster
         )
 
-    # Workaround for a server-side wrapping bug: the frontend's metered handler
-    # drops yarpcerrors.CodeUnavailable on the floor (only CodeDeadlineExceeded
-    # and CodeCancelled are passed through), so FromError re-wraps the plain
-    # fmt.Errorf as CodeUnknown / StatusCode.UNKNOWN.  Detect the scheduler
-    # mid-ContinueAsNew pattern specifically rather than broadening RETRYABLE_CODES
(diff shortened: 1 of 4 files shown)


abhishekj720:
Closing — the root cause was a server-side bug in handler_schedule.go returning yarpcerrors.CodeUnavailable instead of &types.ServiceBusyError. Fixed in cadence-workflow/cadence#8236. Once that merges, RESOURCE_EXHAUSTED flows through the existing RETRYABLE_CODES in RetryInterceptor with no client changes needed.

PR #139, opened 2026-06-16
chore(workflow): move out dataconverter from workflowInstance

shijiesheng:
**What changed?**

* workflowInstance no longer depends on data converter
* data conversion is done directly inside WorkflowEngine

**Why?**

Instead of coupled WE->WI->DC, it's more straightforward to do data conversion inside WE directly. 

**How did you test it?**

Existing Unit Test / Integration Test

**Potential risks**

**Release notes**

**Documentation Changes**


timl3136 (approved):
LGTM

PR #140, opened 2026-06-16
feat: add tag constants and propagate metrics_emitter

timl3136:
**What changed?**
add tag constants and propagate metrics_emitter, also remove constant for features that is yet to be implemented.

**Why?**
First step to fully integrate metrics in operations

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


cadence/metrics/constants.py
neil-xie: are you going to change the metrics names to match with go/jave client like workflow-endtoend-latency -> workflow-endtoend-latency? The names for histogram metrics are timer metrics name + _ns suffix

timl3136: Thanks, that's a good catch. I will make the change and keep it consistent with rest of repos.

neil-xie (approved):

PR #141, opened 2026-06-18
feat: emit poller and worker metrics

timl3136:
**What changed?**

Workers now report **metrics** when they poll Cadence for work and when they start or stop unexpectedly.

- Each **long poll** can increment counters and record how long the poll took. You can see polls that returned a task, polls that returned nothing, and polls that failed (including “might retry” vs “probably won’t”).
- When a task is returned, we can also record **how long the task waited** between being scheduled and starting (when the server gives us the timestamps).
- When the worker process starts its polling loop, we record **worker started** and **pollers started** (one “poller start” per configured poller).
- If the polling loop exits because of an uncaught error, we record a **worker panic** counter.

**Why?**

So people running the Python client can **see what the worker is doing** in dashboards and alerts—similar to other Cadence SDKs—without digging through logs.

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/worker/_decision.py
+++ b/cadence/worker/_decision.py
@@ -63,13 +63,18 @@ class DecisionWorker:
             client, task_list, registry, executor=executor, **options
         )
         self._poller = Poller[PollForDecisionTaskResponse](
-            self._num_pollers, permits, self._poll, self._execute
+            self._num_pollers,
+            permits,
+            self._poll,
+            self._execute,
+            on_start=lambda num_pollers: self._tagged_emitter.counter(
+                POLLER_START_COUNTER, num_pollers
+            ),
         )
         # TODO: Sticky poller, actually running workflows, etc.
 
     async def run(self) -> None:
         self._tagged_emitter.counter(WORKER_START_COUNTER)
-        self._tagged_emitter.counter(POLLER_START_COUNTER, self._num_pollers)
         try:
             await self._poller.run()
         except Exception:
--- a/cadence/worker/_decision.py
+++ b/cadence/worker/_decision.py
@@ -1,5 +1,4 @@
 import asyncio
-import time
 from concurrent.futures import ThreadPoolExecutor
 from typing import Optional
 
@@ -10,7 +9,7 @@ from cadence.api.v1.service_worker_pb2 import (
 from cadence.api.v1.tasklist_pb2 import TaskList, TaskListKind
 from cadence.client import Client
 from cadence.error import CadenceRpcError
-from cadence.metrics import MetricsEmitter, NoOpMetricsEmitter
+from cadence.metrics import MetricsEmitter
 from cadence.metrics.constants import (
     DECISION_POLL_COUNTER,
     DECISION_POLL_FAILED_COUNTER,
@@ -19,14 +18,11 @@ from cadence.metrics.constants import (
     DECISION_POLL_SUCCEED_COUNTER,
     DECISION_POLL_TRANSIENT_FAILED_COUNTER,
     DECISION_SCHEDULED_TO_START_LATENCY,
-    POLLER_START_COUNTER,
     TAG_DOMAIN,
     TAG_TASK_LIST,
-    WORKER_PANIC_COUNTER,
-    WORKER_START_COUNTER,
 )
-from cadence._internal.rpc.retry import RETRYABLE_CODES
 from cadence.worker._decision_task_handler import DecisionTaskHandler
+from cadence.worker._poll_metrics import PollMetrics, run_with_lifecycle_metrics
 from cadence.worker._poller import Poller
 from cadence.worker._registry import Registry
 from cadence.worker._types import _LONG_POLL_TIMEOUT, WorkerOptions
@@ -40,11 +36,20 @@ class DecisionWorker:
         self._task_list = task_list
         self._registry = registry
         self._identity = options["identity"]
-        self._metrics_emitter: MetricsEmitter = options.get(  # type: ignore[attr-defined]
-            "metrics_emitter", NoOpMetricsEmitter()
-        )
+        self._metrics_emitter: MetricsEmitter = options["metrics_emitter"]
         self._tags = {TAG_DOMAIN: client.domain, TAG_TASK_LIST: task_list}
         self._num_pollers = options["decision_task_pollers"]
+        self._poll_metrics = PollMetrics(
+            emitter=self._metrics_emitter,
+            tags=self._tags,
+            poll=DECISION_POLL_COUNTER,
+            latency=DECISION_POLL_LATENCY,
+            succeed=DECISION_POLL_SUCCEED_COUNTER,
+            no_task=DECISION_POLL_NO_TASK_COUNTER,
+            failed=DECISION_POLL_FAILED_COUNTER,
+            transient_failed=DECISION_POLL_TRANSIENT_FAILED_COUNTER,
+            scheduled_to_start=DECISION_SCHEDULED_TO_START_LATENCY,
+        )
         permits = asyncio.Semaphore(
             options["max_concurrent_decision_task_execution_size"]
         )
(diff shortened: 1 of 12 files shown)


cadence/worker/_poll_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+    WORKER_START_COUNTER,
+)
+from cadence.worker._poller import Poller
+
+T = TypeVar("T")
+
+
+@dataclass
+class PollMetrics:
+    emitter: MetricsEmitter
```
natemort: We should update MetricsEmitter to have a with_tags method that creates a new MetricsEmitter with those tags included in the default tags. That way we can avoid passing the tags each time, which is error-prone.

cadence/worker/_poll_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+            self.emitter.counter(self.no_task, tags=self.tags)
+            return
+        self.emitter.counter(self.succeed, tags=self.tags)
+        s, e = task.scheduled_time, task.started_time
+        if s.seconds and e.seconds:
+            lag = (e.seconds + e.nanos / 1e9) - (s.seconds + s.nanos / 1e9)
+            self.emitter.histogram(self.scheduled_to_start, lag, tags=self.tags)
+
+
+async def run_with_lifecycle_metrics(
```
natemort: Can we move this logic(run_with_lifecycle_metrics) into Poller and Worker respectively? There's some error handling in Poller already and it's not clear to me how this is intended to interact with that. I don't know if it's possible for this Exception condition to ever be met.

cadence/worker/_poll_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+                self.latency, time.monotonic() - start, tags=self.tags
+            )
+            metric = self.transient_failed if e.code in RETRYABLE_CODES else self.failed
+            self.emitter.counter(metric, tags=self.tags)
+            raise
+        except Exception:
+            self.emitter.counter(self.failed, tags=self.tags)
+            raise
+
+    def record_result(self, start: float, task: Any) -> None:
```
natemort: We should avoid the use of `Any`. This is relying on the presence of `task_token`, `scheduled_time`, and `started_time` fields. We should introduce a Protocol if we need this to operate on objects with those fields.

cadence/worker/_poll_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+    scheduled_to_start: str
+
+    @contextlib.asynccontextmanager
+    async def track(self) -> AsyncIterator[float]:
+        """Emit poll counter; guarantee exactly one outcome counter on error."""
+        self.emitter.counter(self.poll, tags=self.tags)
+        start = time.monotonic()
+        try:
+            yield start
+        except CadenceRpcError as e:
```
natemort, line 40: Nit: I think I'd prefer for the error handling to be consolidated with the error handling in Poller so that we're logging and emitting metrics for errors in the same place. Handling the errors in multiple places (especially through a context manager) might make it difficult to debug and understand the flow end to end.

cadence/worker/_poll_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+    *,
+    num_pollers: int,
+    tags: dict[str, str],
+) -> None:
+    emitter.counter(WORKER_START_COUNTER, tags=tags)
+    emitter.counter(POLLER_START_COUNTER, num_pollers, tags=tags)
+    try:
+        await poller.run()
+    except Exception:
+        emitter.counter(WORKER_PANIC_COUNTER, tags=tags)
```
natemort: This doesn't match the scenario where we emit the metric in the Go SDK. I think the most similar condition is the handling in Poller for exceptions while executing the task.

Since `Panic` is go specific terminology I'm not sure if it makes sense to emit this metric or a something else.

cadence/worker/_poll_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+
+async def run_with_lifecycle_metrics(
+    emitter: MetricsEmitter,
+    poller: Poller[T],
+    *,
+    num_pollers: int,
+    tags: dict[str, str],
+) -> None:
+    emitter.counter(WORKER_START_COUNTER, tags=tags)
+    emitter.counter(POLLER_START_COUNTER, num_pollers, tags=tags)
```
natemort: Nit: While mathematically equivalent, maybe emitting this from Poller feels slightly more honest? Some change in behavior or bug such that we don't actually start the poller could result in this metric being emitted without reaching the code.

cadence/worker/_decision.py
```diff
@@ -22,6 +35,20 @@ def __init__(
         self._task_list = task_list
         self._registry = registry
         self._identity = options["identity"]
+        self._metrics_emitter: MetricsEmitter = options["metrics_emitter"]
+        self._tags = {TAG_DOMAIN: client.domain, TAG_TASK_LIST: task_list}
+        self._num_pollers = options["decision_task_pollers"]
+        self._poll_metrics = PollMetrics(
```
natemort, line 46: Nit: If we consolidate the tags into the Emitter and pass the Emitter separately from the PollMetrics, we could just have two initialized PollMetrics instances and reference them rather than having the DecisionWorker and ActivityWorker constructors do the initialization.

tests/integration_tests/test_prometheus_metrics.py
```diff
@@ -0,0 +1,164 @@
...
+class WorkflowWithMetrics:
+    @workflow.run
+    async def run(self, message: str) -> str:
+        return await echo_for_metrics.with_options(
+            schedule_to_close_timeout=timedelta(seconds=10)
+        ).execute(message)
+
+
+@contextmanager
+def metrics_endpoint(metrics: PrometheusMetrics) -> Iterator[None]:
```
natemort, line 42: Is this the standard way to expose prometheus metrics?

cadence/worker/_poll_metrics.py
timl3136: Thanks

cadence/worker/_poll_metrics.py
timl3136: Thanks

cadence/worker/_poll_metrics.py
timl3136: Good catch, thanks

cadence/worker/_decision.py
timl3136: Sounds good, i will do this in next PR

tests/integration_tests/test_prometheus_metrics.py
timl3136: not really, what I had earlier is overengineering, I found "start_http_server" from prometheus_client and that is the better way.

cadence/worker/_poll_metrics.py
timl3136: good catch, thanks

cadence/worker/_poll_metrics.py
```diff
@@ -0,0 +1,62 @@
...
+            raise
+
+    def record_result(self, start: float, task: PollTask) -> None:
+        """Record latency, then succeed vs idle, and optional schedule-to-start lag."""
+        self.emitter.histogram(self.latency, time.monotonic() - start)
+        if not (task and task.task_token):
+            self.emitter.counter(self.no_task)
+            return
+        self.emitter.counter(self.succeed)
+        s, e = task.scheduled_time, task.started_time
```
natemort, line 56: Nit: Could we leave these fields as the proto types (instead of introducing a new interface) and use the helper methods to convert them to a timedelta?

natemort (approved):

cadence/worker/_poll_metrics.py
timl3136: Thanks

PR #142, opened 2026-06-24
feat: marker state machine

timl3136:
**What changed?**
Implement marker state machine

**Why?**
Need the state machine for marker feature

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


cadence/_internal/workflow/statemachine/marker_state_machine.py
```diff
@@ -0,0 +1,65 @@
...
+    BaseDecisionStateMachine,
+    DecisionId,
+    DecisionState,
+    DecisionType,
+)
+from cadence._internal.workflow.statemachine.event_dispatcher import EventDispatcher
+from cadence.api.v1 import decision, history
+from cadence.api.v1.common_pb2 import Payload
+
+MARKER_ID_HEADER_KEY = "__cadence_python_marker_id"
```
natemort: How does this compare to the MarkerName in RecordMarkerDecisionAttributes?

I'm not super familiar with Markers, but the Go SDK seems to use that as the ID.

cadence/_internal/workflow/statemachine/nondeterminism.py
```diff
@@ -355,6 +360,38 @@ def _(
...
+@to_expectation.register
+def _(attrs: history.MarkerRecordedEventAttributes) -> Expectation | None:
+    marker_id = marker_id_from_attrs(attrs)
+    if marker_id is None:
+        return None
+    return Expectation(
+        DecisionId(DecisionType.MARKER, marker_id),
+        {
+            "marker_name": attrs.marker_name,
+            "details": attrs.details,
```
natemort: I don't think we need to store the details in the Expectation, I think we'll have to handle it special in DecisionManager. The Expectation exists only for non-determinism tracking.

cadence/_internal/workflow/statemachine/nondeterminism.py
```diff
@@ -355,6 +360,38 @@ def _(
...
 
+# Markers - Enforce SDK marker id and marker name. Details are recorded data, so
+# replay returns the history value instead of requiring workflow code to recreate it.
+#
+# 'details' is intentionally absent from the decision-side handler: Expectation.__eq__
+# only compares keys present in both sides, so omitting it here lets the match succeed
+# on name+id alone. The history-side handler (below) carries 'details' so DecisionManager
+# can retrieve the recorded value and return it to the caller on replay.
+@to_expectation.register
+def _(attrs: decision.RecordMarkerDecisionAttributes) -> Expectation | None:
```
natemort, line 376: The Go SDK (haven't looked at Java) names markers with as %s_%s, where the first parameter is the type of the marker (local activity, side effect, version, etc.) and the second is the contextual identifier for that marker (activity ID, side effect ID, version marker). I think our state machine will unfortunately have to do something similar and behave differently depending on the type. We could even consider a different state machine for each, but LocalActivities is really the only one that's massively different.

Adding a new Version Marker for example is totally fine and we don't want to create an Expectation for either the decision or the event.

Adding a new LocalActivity on the other hand should be flagged on nondeterminism. I think side effects are similar.

For LocalActivities we'll also have to do immediate cancellation and expose a Future and everything, but we can ignore that for now.

cadence/_internal/workflow/statemachine/nondeterminism.py
```diff
@@ -128,7 +131,7 @@ def validate_cancel(self, decision_id: DecisionId) -> None:
             Expectation(decision_id=decision_id, properties=CANCEL)
         )
 
-    def _validate_expectation(self, actual: Expectation) -> None:
+    def _validate_expectation(self, actual: Expectation) -> Expectation:
```
natemort, line 136: Nit: Can you document the new return value?

cadence/_internal/workflow/statemachine/nondeterminism.py
timl3136: Thanks, I removed details from both handlers.

cadence/_internal/workflow/statemachine/nondeterminism.py
timl3136: Added comments for checking expectation.

cadence/_internal/workflow/statemachine/marker_state_machine.py
timl3136: I checked with Go SDK, and I added a permalink for the formatting "%v_%v" in the comments

cadence/_internal/workflow/statemachine/nondeterminism.py
timl3136: Thanks, I have made those changes.

cadence/_internal/workflow/statemachine/marker_state_machine.py
```diff
@@ -0,0 +1,91 @@
...
+marker_events = EventDispatcher()
+
+
+def encode_marker_details(context_id: str, user_data: bytes) -> bytes:
+    """Encode [context_id, user_data] into a single bytes value stored in Details.
+
+    Format: 4-byte big-endian length of context_id, then context_id UTF-8, then user_data.
+    """
+    ctx = context_id.encode("utf-8")
+    return len(ctx).to_bytes(4, "big") + ctx + user_data
```
natemort: I think the other SDKs align on this at least being a JSON object, let's do the same rather than introducing a more complex encoding.

I think we can punt that to a subsequent PR as well. For the state machine and non-determinism I don't think we really care what specifically is in the Marker, just the type (which we infer from the name).

When adding LocalActivities or other things we might alter the non-determinism

cadence/_internal/workflow/statemachine/marker_state_machine.py
```diff
@@ -0,0 +1,91 @@
...
+    Returns (None, data) if the encoding is absent or malformed (e.g. markers from
+    other SDKs or pre-encoding history).
+    """
+    if len(data) < 4:
+        return None, data
+    ctx_len = int.from_bytes(data[:4], "big")
+    if 4 + ctx_len > len(data):
+        return None, data
+    try:
+        context_id = data[4 : 4 + ctx_len].decode("utf-8")
```
natemort: We can parse this from the marker name rather than the data.

cadence/_internal/workflow/statemachine/marker_state_machine.py
```diff
@@ -0,0 +1,91 @@
...
+from cadence.api.v1 import decision, history
+
+# Marker type names match the Go SDK constants:
+# https://github.com/cadence-workflow/cadence-go-client/blob/727b555be0fd0f65ad201832ba078b661919034e/internal/internal_decision_state_machine.go#L160-L163
+SIDE_EFFECT_MARKER_NAME = "SideEffect"
+VERSION_MARKER_NAME = "Version"
+LOCAL_ACTIVITY_MARKER_NAME = "LocalActivity"
+MUTABLE_SIDE_EFFECT_MARKER_NAME = "MutableSideEffect"
+
+KNOWN_MARKER_NAMES = frozenset(
```
natemort, line 26: One additional one is the immediate cancellation marker `Cancel_` which is unique to Python.

cadence/_internal/workflow/statemachine/nondeterminism.py
```diff
@@ -355,6 +362,39 @@ def _(
...
+@to_expectation.register
+def _(attrs: decision.RecordMarkerDecisionAttributes) -> Expectation | None:
+    context_id, _ = decode_marker_details(attrs.details.data)
+    if context_id is None:
+        return None
+    if attrs.marker_name == VERSION_MARKER_NAME:
+        return None
+    return Expectation(
+        DecisionId(DecisionType.MARKER, f"{attrs.marker_name}_{context_id}"),
+        {"marker_name": attrs.marker_name},
```
natemort: We can have an empty expectation body, just the correct DecisionID is enough for now. That covers type and name. When we add LocalActivities and other things with more parameters we might introduce additional checks, but we want to be really specifi with them.

cadence/_internal/workflow/statemachine/marker_state_machine.py
timl3136: Thanks, they are now encoded as a json object.

cadence/_internal/workflow/statemachine/marker_state_machine.py
timl3136: since we encode them as json object, for decoding part, we can just json.decode to get the context_id and data

cadence/_internal/workflow/statemachine/marker_state_machine.py
timl3136: I added a cancel bypass in decision_manager.

cadence/_internal/workflow/statemachine/nondeterminism.py
timl3136: Addressed.

shijiesheng (approved):

PR #145, opened 2026-07-07
feat(metrics): emit workflow lifecycle metrics

timl3136:
**What changed?**
Added workflow and activity lifecycle metrics across the worker execution path:

**Why?**

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


cadence/worker/_decision_task_handler.py
```diff
@@ -154,18 +183,38 @@ async def _handle_task_implementation(
...
+                workflow_engine.process_decision,
+                workflow_events,
+                task.query if is_query_task else None,
+            )
+        except Exception:
+            emitter.counter(DECISION_EXECUTION_FAILED_COUNTER)
+            emitter.counter(DECISION_TASK_PANIC_COUNTER)
+            raise
+        finally:
+            emitter.histogram(
```
natemort, line 201: Is there some way we can change this interface to simplify its usage?

The mismatch of units between time.monotonic() and this parameter seems error prone. It would be really easy to pass a value that's off by a factor of 1e9.

Can we pass a more specific type rather than a number as the parameter? Can we use some sort of stopwatch object that we start/stop to report latency?

cadence/_internal/activity/_activity_executor.py
```diff
@@ -44,13 +58,38 @@ def __init__(
...
+        wf_type = task.workflow_type.name if task.workflow_type else ""
+        emitter = self._metrics_emitter.with_tags(
+            {
+                TAG_ACTIVITY_TYPE: activity_type,
+                TAG_WORKFLOW_TYPE: wf_type,
+                TAG_DOMAIN: self._client.domain,
+                TAG_TASK_LIST: self._task_list,
+            }
+        )
+        scheduled_ts = task.scheduled_time.seconds + task.scheduled_time.nanos / 1e9
```
natemort: Is there some helper we can use for this? Alternatively can we introduce one? It's a pretty common operation

cadence/_internal/activity/_activity_executor.py
```diff
@@ -82,8 +121,14 @@ def _create_context(
             )
 
     async def _report_failure(
-        self, task: PollForActivityTaskResponse, error: Exception
+        self,
+        task: PollForActivityTaskResponse,
+        error: Exception,
+        emitter: Optional[MetricsEmitter] = None,
+        e2e_latency: Optional[float] = None,
```
natemort: Why are these two fields optional?

cadence/_internal/activity/_activity_executor.py
timl3136: Thanks, I will create a new PR for those to simplify metrics collection and rebased this branch later

cadence/worker/_decision_task_handler.py
timl3136: I wrote a helper function to address it here: https://github.com/cadence-workflow/cadence-python-client/pull/149

cadence/_internal/activity/_activity_executor.py
timl3136: I made both non-optional, but we would not emit if the e2e_latency is <0 (which should never happen)

cadence/_internal/workflow/history_event_iterator.py
```diff
@@ -1,15 +1,28 @@
...
+
+if TYPE_CHECKING:
+    from cadence.metrics import MetricsEmitter
 
 
 async def iterate_history_events(
-    decision_task: PollForDecisionTaskResponse, client: Client
+    decision_task: PollForDecisionTaskResponse,
+    client: Client,
+    metrics_emitter: Optional[MetricsEmitter] = None,
```
natemort, line 25: Nit: This shouldn't be Optional. If it's possible for the caller not to have one, the caller should have a Noop implementation.

cadence/_internal/workflow/history_event_iterator.py
```diff
@@ -1,15 +1,28 @@
...
     GetWorkflowExecutionHistoryRequest,
     GetWorkflowExecutionHistoryResponse,
 )
 from cadence.client import Client
+from cadence.metrics.constants import (
+    WORKFLOW_GET_HISTORY_COUNTER,
+    WORKFLOW_GET_HISTORY_LATENCY,
+)
+
+if TYPE_CHECKING:
```
natemort, line 18: Nit: Is this needed? It seems unlikely that we have a circular dependency between these.

cadence/worker/_decision_task_handler.py
```diff
@@ -154,18 +185,38 @@ async def _handle_task_implementation(
...
+                DECISION_EXECUTION_LATENCY, time.monotonic_ns() - exec_start_ns
+            )
         if is_query_task:
             if not decision_result.query_result:
                 raise ValueError("Query result is empty")
             await self._respond_query_task_completed(task, decision_result.query_result)
         else:
-            await self._respond_decision_task_completed(task, decision_result)
+            await self._respond_decision_task_completed(task, decision_result, emitter)
+            outcome = next(
```
natemort, line 210: Nit: this is complicated, I think it would be easier to read if we passed the entire set of decisions  to the helper.

cadence/worker/_decision_task_handler.py
```diff
@@ -262,8 +313,30 @@ async def handle_task_failure(
                 exc_info=True,
             )
 
+    def _emit_workflow_outcome_metrics(
+        self, outcome: str, workflow_events: list, emitter: MetricsEmitter
+    ) -> None:
+        counter_map = {
+            "completed": WORKFLOW_COMPLETED_COUNTER,
```
natemort, line 320: Nit: Introduce constants for these and a type alias to ensure they stay consistent between this method and the outcome.

cadence/worker/_decision_task_handler.py
```diff
@@ -262,8 +313,30 @@ async def handle_task_failure(
...
+        }
+        if counter := counter_map.get(outcome):
+            emitter.counter(counter)
+
+        if workflow_events:
+            now = Timestamp()
+            now.GetCurrentTime()
+            e2e_latency_ns = duration_between_ns(workflow_events[0].event_time, now)
+            if e2e_latency_ns is not None and e2e_latency_ns >= 0:
+                emitter.histogram(WORKFLOW_END_TO_END_LATENCY, e2e_latency_ns)
```
natemort, line 333: Nit: In a followup PR, can we change MetricsEmitter.histogram from accepting a float to a Duration? That way it's impossible to get the unit wrong.

cadence/worker/_decision_task_handler.py
```diff
@@ -262,8 +313,30 @@ async def handle_task_failure(
                 exc_info=True,
             )
 
+    def _emit_workflow_outcome_metrics(
+        self, outcome: str, workflow_events: list, emitter: MetricsEmitter
```
natemort, line 317: Nit: Pass the workflow duration explicitly rather than the workflow_events.

natemort (approved):

PR #156, opened 2026-07-28
feat: mutable side_effect feature

timl3136:
**What changed?**
Added mutable side-effect support, including marker metadata, replay caching, and public workflow APIs.
Mutable side-effect markers now use stable IDs derived from the side-effect ID and access count, rather than consuming the shared decision-ID sequence.

**Why?**
Mutable side effects record a value only when it changes while returning recorded values during replay. It allows use of function like random number without triggering non-determinism issue.

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/workflow/context.py
+++ b/cadence/_internal/workflow/context.py
@@ -285,6 +285,54 @@ class Context(WorkflowContext):
             self.data_converter().from_data(result_payload, [result_type])[0],
         )
 
+    def mutable_side_effect(
+        self,
+        id: str,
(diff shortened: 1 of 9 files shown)


shijiesheng (approved):
unrelated to this change. The two lines are redundant. https://github.com/cadence-workflow/cadence-python-client/blob/b5ff143651605526af4fbcaefe803b4adb4b3ddf/cadence/_internal/workflow/statemachine/decision_manager.py#L441

PR #158, opened 2026-07-29
feat: propagate context through workflow execution

timl3136:
**What changed?**
Context propagation

**Why?**
Necessary feature

**How did you test it?**
Unit and integration tests

**Potential risks**

**Release notes**

**Documentation Changes**

related issue: https://github.com/cadence-workflow/cadence-python-client/issues/160

### DIFF
--- a/cadence/__init__.py
+++ b/cadence/__init__.py
@@ -6,6 +6,7 @@ A Python framework for authoring workflows and activities for Cadence.
 
 # Import main client functionality
 from .client import Client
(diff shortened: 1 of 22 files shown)


shijiesheng:
I don't see places where we set the headers before start workflows.

shijiesheng:
Because it's in the inject() method, I read it wrong.

shijiesheng (approved):

PR #163, opened 2026-08-05
feat: implement workflow versioning support

timl3136:
**What changed?**
Added support for workflow versioning

**Why?**
Important feature now that we have marker state machine

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cadence/_internal/workflow/context.py
+++ b/cadence/_internal/workflow/context.py
@@ -17,8 +17,7 @@ from cadence._internal.workflow.statemachine.marker_state_machine import (
 from cadence._internal.workflow.versioning import (
     decode_version_marker_details,
     encode_version_marker_details,
-    select_version,
-    validate_selected_version,
+    validate_resolved_version,
     validate_version_arguments,
 )
 from cadence._internal.context import inject_headers, set_header
@@ -47,7 +46,6 @@ from cadence.workflow import (
     WorkflowCancellationInfo,
     WorkflowContext,
     WorkflowInfo,
-    VersioningOption,
     DEFAULT_VERSION,
 )
 from cadence.api.v1.history_pb2 import WorkflowExecutionCancelRequestedEventAttributes
@@ -71,9 +69,6 @@ class Context(WorkflowContext):
         self._decision_manager = decision_manager
         self._context_propagators = context_propagators
         self._cancellation_info: WorkflowCancellationInfo | None = None
-        # None represents a provisional DEFAULT_VERSION returned while replaying
-        # history before a Version marker becomes available.
-        self._versions: dict[str, int | None] = {}
 
     def info(self) -> WorkflowInfo:
         return self._info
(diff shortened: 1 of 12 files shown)


cadence/_internal/workflow/context.py
```diff
@@ -338,6 +350,128 @@ def mutable_side_effect(
...
+        if version < min_supported or version > max_supported:
+            raise FatalDecisionError(
+                f"{source} version {version} for change_id {change_id!r} is outside "
+                f"the supported range [{min_supported}, {max_supported}]"
+            )
+
+    def _decode_recorded_version(self, change_id: str, details: Payload) -> int:
+        # Subclasses may override serialization, so only the canonical converter
+        # is guaranteed to encode Version marker details as a single JSON value.
+        if type(self.data_converter()) is DefaultDataConverter:
```
natemort: We shouldn't use the user provided DataConverter for handling internal types. They could just as easily fail to serialize our type.

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -282,6 +287,29 @@ def record_mutable_side_effect(
         state.value = result
         return result
 
+    def version_marker_details(self, change_id: str) -> Payload | None:
+        """Return a preloaded Python Version marker's details, if one exists."""
+        return self._recorded_marker_details.get(
+            marker_decision_id(VERSION_MARKER_NAME, change_id)
+        )
+
+    def has_pending_version_marker(self, change_id: str) -> bool:
```
natemort: It's not clear to me why we need to track the event IDs. Are we processing the history events for markers more than once?

cadence/_internal/workflow/statemachine/decision_manager.py
```diff
@@ -439,7 +514,9 @@ def _start_execution(self, replaying: bool, outcomes: List[history.HistoryEvent]
         for event in outcomes:
             self._determinism_tracker.add_expectation(event)
             if event.HasField("marker_recorded_event_attributes"):
-                self._index_marker_details(event.marker_recorded_event_attributes)
```
natemort, line 442: I think I missed this previously, why do we index markers as part of nondeterminism? Don't we already preload them separately?

tests/integration_tests/workflow/test_versioning.py
```diff
@@ -0,0 +1,126 @@
...
+async def test_version_marker_is_recorded_and_replayed_by_cadence(
+    helper: CadenceHelper,
+) -> None:
+    async with helper.worker(registry) as worker:
+        execution = await worker.client.start_workflow(
+            "VersioningWorkflow",
+            task_list=worker.task_list,
+            execution_start_to_close_timeout=timedelta(seconds=10),
+        )
+        # This ensures release is handled in a later decision task. The worker
```
natemort: I'm not sure I understand this comment. Sticky caching shouldn't influence behavior.

tests/integration_tests/workflow/test_versioning.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(self) -> None:
+        self._released = False
+
+    @workflow.signal(name="release")
+    def release(self) -> None:
+        self._released = True
+
+    @workflow.run
+    async def run(self) -> str:
+        from_marker = workflow.get_version(_CHANGE_ID, 1, 2)
```
natemort: Can we add some more rigorous test scenarios that involve replaying workflow histories, similar to the `TestVersionedWorkflowV X` test scenarios in the Go SDK? I don't think this sufficiently demonstrates versioning and the expected behavior in different scenarios.

cadence/_internal/workflow/versioning.py
```diff
@@ -0,0 +1,47 @@
...
+def validate_version_arguments(
+    change_id: str, min_supported: int, max_supported: int
+) -> None:
+    if not isinstance(change_id, str) or not change_id:
+        raise ValueError("change_id must be a non-empty str")
+    for name, value in (
+        ("min_supported", min_supported),
+        ("max_supported", max_supported),
+    ):
+        if isinstance(value, bool) or not isinstance(value, int):
```
natemort, line 40: Why are we checking if it's a bool?

cadence/_internal/workflow/context.py
```diff
@@ -338,6 +350,128 @@ def mutable_side_effect(
...
+        # code path. There is no marker to validate or state machine to create.
+        if self.is_replay_mode():
+            self._validate_recorded_version(
+                change_id,
+                DEFAULT_VERSION,
+                min_supported,
+                max_supported,
+                "markerless replay",
+            )
+            self._versions[change_id] = None
```
natemort: This introduces a bug. If I were to call get_version for a ChangeID twice (which is allowed and happens when versioning is within a loop), the second time `_versions` would have a value of `None` and result in the evaluation logic at the top of the function being executed.

cadence/_internal/workflow/context.py
```diff
@@ -61,6 +70,9 @@ def __init__(
         self._decision_manager = decision_manager
         self._context_propagators = context_propagators
         self._cancellation_info: WorkflowCancellationInfo | None = None
+        # None represents a provisional DEFAULT_VERSION returned while replaying
+        # history before a Version marker becomes available.
+        self._versions: dict[str, int | None] = {}
```
natemort: What's the benefit of maintaining this versions dict here vs relying on the state machine?

It seems like caching this is incorrect. If I were to call get_version before we've processed a HistoryEvent that contains the version marker then we'd cache a value of DEFAULT_VERSION. When we later process the HistoryEvent this cache would never be updated.

cadence/_internal/workflow/context.py
```diff
@@ -338,6 +350,128 @@ def mutable_side_effect(
...
+
+        selected = select_version(min_supported, max_supported, *options)
+        validate_selected_version(change_id, selected, min_supported, max_supported)
+        self._versions[change_id] = selected
+
+        # DEFAULT_VERSION is intentionally never written to history.
+        if selected == DEFAULT_VERSION:
+            return selected
+
+        self._decision_manager.record_version_marker(
```
natemort: The other clients also add SearchAttributes, will we do the same?

cadence/_internal/workflow/context.py
timl3136: Good call, i will remove it. Thanks for flagging.

cadence/_internal/workflow/statemachine/decision_manager.py
timl3136: You are right, this is unnecessary.

cadence/_internal/workflow/statemachine/decision_manager.py
timl3136: I forget to remove this logic when I added the preload, I will remove it now.

cadence/_internal/workflow/versioning.py
timl3136: because bool is subclass of int, if someone pass in True/False, it will be interpret as a int for isinstance.

tests/integration_tests/workflow/test_versioning.py
timl3136: I am not entirely sure once we have sticky cache, how should we test this behavior.

cadence/_internal/workflow/context.py
timl3136: Thanks for reminding, I will add search attribute in a later pr

cadence/_internal/workflow/context.py
timl3136: marker state machine only tracks a recordMarker decision. For the DEFAULT_VERSION, we do not create marker state machine for that. I think adding the None here caused some confusion, and I will make that explicit.

cadence/_internal/workflow/context.py
timl3136: Like the comment below, I will remove the overload on None.

tests/integration_tests/workflow/test_versioning.py
timl3136: Got it, i will add those.

shijiesheng (approved):

